PR #112, opened 2011-11-22
Allow for OAuth 1 auth by passing OAuth::AccessToken

Kimtaro:
Added support for OAuth 1 signing when passing an OAuth::AccessToken as options[:oauth_auth].

But obtaining the access token is up to the developer.

Cheers
Kim

### DIFF
--- a/examples/twitter_oauth.rb
+++ b/examples/twitter_oauth.rb
@@ -1,0 +1,31 @@
+dir = File.expand_path(File.join(File.dirname(__FILE__), '..', 'lib'))
+require File.join(dir, 'httparty')
+require 'oauth'
+require 'pp'
+config = YAML::load(File.read(File.join(ENV['HOME'], '.twitter_oauth')))
+
+class Twitter
+  include HTTParty
+  base_uri 'twitter.com'
+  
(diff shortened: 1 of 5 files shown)


christoph-buente:
I am using httparty for https://github.com/sozialhelden/rosemary. This is a library to talk to the OpenStreetMap API. They offer BasicAuth and OAuth requests. Up to httparty v0.11.0 it worked fine. 0.12 made changes that break my code. I have to do a lot of stunts to distinguish between BasicAuth and OAuth requests and handle them accordingly.

In times where "login with this service" type of requests are becoming more common i think it is a scope of a http library to cater this need. The oauth gem did not update since 2012 as the specification did not change. It was replaced by OAuth2, which is another library.

Could you reconcider pulling in this PR?

Cheers, Christoph

jnunemaker:
@christoph-buente sorry, but I'm sticking to my previous comment. If there is anything in httparty that is preventing this from being bolted on top (composed with httparty) to support oauth, I'm interested, but I'm not interested in requiring the oauth gem and supporting oauth myself going forward. I'm not the right person for that.

PR #139, opened 2012-05-07
Remove goofy post install message.

raykrueger:
This post install message simply adds noise to our lives. Do we really need more noise?
I don't think so :)

### DIFF
--- a/httparty.gemspec
+++ b/httparty.gemspec
@@ -15,8 +15,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'multi_json', "~> 1.0"
   s.add_dependency 'multi_xml'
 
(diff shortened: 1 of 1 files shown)


michaelglass:
easy enough to keep a fork called `HTTPolite`, right? Me? I'm in the Pro Party Party.

danlo:
This is kind of like an HTTP 418 error.  It's kind of like an insider joke, like when I'm looking at some other sites javascript code and I see a link/job application for an engineer or.. like giving the standard response to the standard password "12345" 

https://www.youtube.com/watch?v=a6iW-8xPw3k  .... Spaceballs reference to luggage.

PR #211, opened 2013-05-27
When you HTTParty, try to keep the noise down

tobico:


nixpulvis:
> But who uses gem anyway.

I sure do!

PR #236, opened 2013-09-23
Remove post_install_message from gemspec

eric:
It's cute but it ends up often just being extra noise when installing gems.


oskarpearson:
How to disable the post-install message: 

https://github.com/jnunemaker/httparty/pull/321#issuecomment-525375166

PR #315, opened 2014-07-29
Party more quietly

avdi:
OK, I think after all these years we have established that the post-install message works. Now can we please stop cluttering up install output? People are trying to sleep over here!


oskarpearson:
https://github.com/jnunemaker/httparty/pull/321#issuecomment-525375166

PR #365, opened 2014-12-30
Replace URI.encode with ERB::Util.url_encode

mikeastock:
URI.encode is an alias for URI.escape which has been deprecated. There
is a great stackoverflow question discussing this deprecation and
possible replacements, http://stackoverflow.com/a/2832003/1769125. 

After some testing I decided that ERB::Util.url_encode was the best choice as
it most closely mirrored the current implementation.

```ruby
foo = "http://google.com?query=hello"

old_way = URI.escape(foo, Regexp.new("[^#{URI::PATTERN::UNRESERVED}]"))
#=> "http%3A%2F%2Fgoogle.com%3Fquery%3Dhello"

new_way = ERB::Util.url_encode(foo)
#=> "http%3A%2F%2Fgoogle.com%3Fquery%3Dhello"

old_way == new_way
#=> true
```

### DIFF
--- a/lib/httparty/hash_conversions.rb
+++ b/lib/httparty/hash_conversions.rb
@@ -30,7 +30,7 @@ module HTTParty
       elsif value.respond_to?(:to_hash)
         stack << [key,value.to_hash]
       else
-        param << "#{key}=#{URI.encode(value.to_s, Regexp.new("[^#{URI::PATTERN::UNRESERVED}]"))}&"
+        param << "#{key}=#{ERB::Util.url_encode(value.to_s)}&"
       end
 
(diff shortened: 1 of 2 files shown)


jnunemaker:
@mikeastock tests would definitely be a good idea. I guess what I'm not clear on is why to use ERB over CGI for this.

mikeastock:
@jnunemaker the reason to go with ERB over CGI is that ERB::Util.url_encode properly uses %20 for spaces.

```ruby
CGI.escape("Mike Stock")
#=> "Mike+Stock"
ERB::Util.url_encode("Mike Stock")
#=> "Mike%20Stock"
```

PR #367, opened 2015-01-01
Create spec for HashConversion

mikeastock:
Create a basic spec for HashConversion


rusikf:
@mikeastock , [i added pr to your branch](https://github.com/mikeastock/httparty/pull/2), what you think about it?
Adding HTTParty module(used everywhere as namespace) only once for more consice spec.

PR #371, opened 2015-01-20
remote_ip to response

rusikf:
I want to implement getting remote ip for response host like this:
```ruby
response = HTTParty.get 'http://google.ru'
response.remote_ip
```
```shell
=> "173.194.40.248"
```
I don't find this ability in net/http lib, only in socket library.
I am also, don't created tests, because i am not sure, that my implementation looks clearly.
@jnunemaker , what you think about it ?

### DIFF
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -6,6 +6,7 @@ require 'zlib'
 require 'multi_xml'
 require 'json'
 require 'csv'
+require 'socket'
 
 require 'httparty/module_inheritable_attributes'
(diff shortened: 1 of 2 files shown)


jnunemaker:
I'm not sure this makes sense in HTTParty. I'd probably keep this in application code. Is there anything in HTTParty that we would have to expose to make that possible?

rusikf:
@jnunemaker , I took this idea from excon gem. It uses sockets internally and have some additional features like this (remote_ip)

PR #395, opened 2015-05-08
Replace Ruby's URI with Addressable::URI

### LINKED ISSUE #394: http does not accept registry part
Hi there!

I'm using HTTParty in a project involved with podcasts management and found a problem we don't know how to solve but using another http requester (which we don't want).

Requesting this URL (http://svd_identity.spinninpodcasts.com/rss) raises this error:
```
ruby-2.0.0-p353/lib/ruby/2.0.0/uri/generic.rb:214:in `initialize': the scheme http does not accept registry part: svd_identity.spinninpodcasts.com (or bad hostname?) (URI::InvalidURIError)
```
[Here](http://stackoverflow.com/questions/5208851/is-there-a-workaround-to-open-urls-containing-underscores-in-ruby) they say it's a problem with the use of _URI.parse_.
Does HTTParty have any alternative?

Thanks.

jbgutierrez:
Addressable is a replacement for the URI implementation that is part of
Ruby's standard library. It more closely conforms to RFC 3986, RFC 3987,
and RFC 6570 (level 4), additionally providing support for IRIs and URI
templates. http://addressable.rubyforge.org/

Fixes #394

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ gemspec
 gem 'rake'
 gem 'fakeweb',  '~> 1.3'
 gem 'mongrel',  '1.2.0.pre2'
+gem 'addressable'
 
 group :development do
   gem 'guard'
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -1,7 +1,7 @@
 require 'pathname'
 require 'net/http'
 require 'net/https'
-require 'uri'
+require "addressable/uri"
 require 'zlib'
 require 'multi_xml'
 require 'json'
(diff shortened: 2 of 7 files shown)


jnunemaker:
I'm not interested in forcing use of addressable. I like to keep dependencies slim. I was more thinking that we make uri parsing pluggable with the default being URI.parse. Does that make sense?

jbgutierrez:
It totally makes sense,
I'll try something similar to this:

```ruby
class Foo
  include HTTParty
  uri_adapter Addressable::URI
end
```

jnunemaker:
That works.

PR #402, opened 2015-05-14
Allow setting a custom URI adapter

### LINKED ISSUE #394: http does not accept registry part

jbgutierrez:
This option allows you to roll out fancier URI implementations like the one
offered by the Addressable gem.

Addressable is a replacement for the URI implementation that is part of
Ruby's standard library. It more closely conforms to RFC 3986, RFC 3987,
and RFC 6570 (level 4), additionally providing support for IRIs and URI
templates. http://addressable.rubyforge.org/

*Sample Usage:*

```javascript
require 'httparty'
require 'addressable/uri'
require 'pp'

class Foo
  include HTTParty
  uri_adapter Addressable::URI
end

pp Foo.get('http://svd_identity.spinninpodcasts.com/rss')
```

Fixes #394

### DIFF
--- a/lib/httparty/connection_adapter.rb
+++ b/lib/httparty/connection_adapter.rb
@@ -69,10 +69,11 @@ module HTTParty
 
     def connection
       host = clean_host(uri.host)
+      default_port = uri.scheme == 'https' ? 443 : 80
       if options[:http_proxyaddr]
-        http = Net::HTTP.new(host, uri.port, options[:http_proxyaddr], options[:http_proxyport], options[:http_proxyuser], options[:http_proxypass])
+        http = Net::HTTP.new(host, uri.port || default_port, options[:http_proxyaddr], options[:http_proxyport], options[:http_proxyuser], options[:http_proxypass])
       else
-        http = Net::HTTP.new(host, uri.port)
+        http = Net::HTTP.new(host, uri.port || default_port)
       end
 
       http.use_ssl = ssl_implied?(uri)
@@ -133,7 +134,7 @@ module HTTParty
     end
 
     def ssl_implied?(uri)
-      uri.port == 443 || uri.instance_of?(URI::HTTPS)
+      uri.port == 443 || uri.scheme == 'https'
     end
 
     def attach_ssl_certificates(http, options)
--- a/lib/httparty/connection_adapter.rb
+++ b/lib/httparty/connection_adapter.rb
@@ -70,7 +70,7 @@ module HTTParty
 
     def connection
       host = clean_host(uri.host)
-      port = uri.port or uri.scheme == 'https' ? 443 : 80
+      port = uri.port || (uri.scheme == 'https' ? 443 : 80)
       if options[:http_proxyaddr]
         http = Net::HTTP.new(host, port, options[:http_proxyaddr], options[:http_proxyport], options[:http_proxyuser], options[:http_proxypass])
       else
--- a/lib/httparty/connection_adapter.rb
+++ b/lib/httparty/connection_adapter.rb
@@ -70,11 +70,11 @@ module HTTParty
 
     def connection
       host = clean_host(uri.host)
-      default_port = uri.scheme == 'https' ? 443 : 80
+      port = uri.port or uri.scheme == 'https' ? 443 : 80
       if options[:http_proxyaddr]
-        http = Net::HTTP.new(host, uri.port || default_port, options[:http_proxyaddr], options[:http_proxyport], options[:http_proxyuser], options[:http_proxypass])
+        http = Net::HTTP.new(host, port, options[:http_proxyaddr], options[:http_proxyport], options[:http_proxyuser], options[:http_proxypass])
       else
-        http = Net::HTTP.new(host, uri.port || default_port)
+        http = Net::HTTP.new(host, port)
       end
 
       http.use_ssl = ssl_implied?(uri)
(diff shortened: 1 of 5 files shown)


lib/httparty/connection_adapter.rb
```diff
@@ -69,10 +70,11 @@ def initialize(uri, options = {})
 
     def connection
       host = clean_host(uri.host)
+      default_port = uri.scheme == 'https' ? 443 : 80
```
jnunemaker: This should be uri.port. Why the change?

lib/httparty/request.rb
```diff
@@ -89,6 +90,19 @@ def parser
       options[:parser]
     end
 
+    def URI uri
```
jnunemaker: Why this method? Why passing uri to it? I don't get this.

lib/httparty/connection_adapter.rb
jbgutierrez: Well, I tried to decouple from URI in the first commit and this was a biased change.
Addressable:URI sets port to nil when the url is pointing to a default port. Ruby's URI [sets it to 80/443](https://github.com/ruby/ruby/blob/trunk/lib/uri/generic.rb#L210) so that's why I've set the same fallback:

```ruby
require "minitest/autorun"
require "uri"
require "addressable/uri"

class TestRubyImplementations < Minitest::Unit::TestCase

  def test_uri_port_defaults
    url = 'http://foo/bar'
    uri = URI.parse url
    assert_equal uri.port, 80

    uri = Addressable::URI.parse url
    assert_equal uri.port, nil
  end

end
```

lib/httparty/request.rb
jbgutierrez: `httparty` uses [URI's kernel method](https://github.com/ruby/ruby/blob/trunk/lib/uri/common.rb#L709) which is tightly coupled to Ruby `URI.parse` method. I thought the smallest discrete change that would help to introduce URI adaptor was to hide it from the `httparty` module. On a third commit I've inlined the method to make that more clear.

lib/httparty/connection_adapter.rb
jnunemaker: The issue is that sometimes people run https on non-443 port. I've had people complain when I assume the port in the past.

jbgutierrez: Definitely and the current source covers that.

While Ruby's URI ensures a value on the port attribute after its initialization, Addressable's URI doesn't set it when parsing a URL running on a default port (no mather it's https or http). That's why I've set this fallback.

```Ruby
require "minitest/autorun"
require "uri"
require "addressable/uri"

class TestRubyImplementations < Minitest::Unit::TestCase

  def test_uri_parse_on_default_ports
    url = 'https://foo/bar'
    uri = URI.parse url
    assert_equal uri.port, 443

    uri = Addressable::URI.parse url
    assert_equal uri.port, nil
  end

  def test_uri_parse_on_non_default_ports
    url = 'https://foo:8443/bar'
    uri = URI.parse url
    assert_equal uri.port, 8443

    uri = Addressable::URI.parse url
    assert_equal uri.port, 8443
  end

end
```
Hoping to make it clear, I'm DRYing up default port fallback.

lib/httparty/request.rb
```diff
@@ -12,7 +12,7 @@ class Request #:nodoc:
       Net::HTTP::Copy
     ]
 
-    SupportedURISchemes  = [URI::HTTP, URI::HTTPS, URI::Generic]
+    SupportedURISchemes  = ['http', 'https', nil]
```
jnunemaker: nil is confusing here but I don't have a better idea.

lib/httparty/request.rb
```diff
@@ -31,20 +31,30 @@ class Request #:nodoc:
...
       set_basic_auth_from_uri
     end
 
     def path=(uri)
-      @path = URI(uri)
+      uri_adapter = options[:uri_adapter]
+
+      @path = if uri.is_a?(uri_adapter)
+        uri
+      elsif String.try_convert(uri)
```
jnunemaker: TIL. Honestly, I've never heard of this. Part of me would rather have the error raise if to_str is not a method on uri (ie: uri.to_str), but I'm not sure really.

jbgutierrez: But that's not exactly what it's doing right now. I mean might break backwards compatibility, right?

lib/httparty/request.rb
jbgutierrez: I'm starting to have second thoughts on this and I've detected a false PASS Webcal URI test. In the end, all requests are done through HTTPS o HTTP no matter which scheme is originally specify.

Wouldn't it make sense to remove this supported URI schemes list? It all started with this [pull request](https://github.com/jnunemaker/httparty/pull/164).

jnunemaker:
All good to go on this?

jbgutierrez:
Yep!, Looks good to me!

PR #408, opened 2015-06-10
Custom loggers

barberj:
Enable adding custom loggers to available formatters.

### DIFF
--- a/lib/httparty/logger/logger.rb
+++ b/lib/httparty/logger/logger.rb
@@ -3,16 +3,19 @@ require 'httparty/logger/curl_logger'
 
 module HTTParty
   module Logger
+    def self.formatters
+      {
+        :curl => Logger::CurlLogger,
+        :apache => Logger::ApacheLogger
+      }
+    end
+
     def self.build(logger, level, formatter)
       level ||= :info
       formatter ||= :apache
 
-      case formatter
-      when :curl
-        Logger::CurlLogger.new(logger, level)
-      else
-        Logger::ApacheLogger.new(logger, level)
-      end
+      logger_klass = formatters[formatter] || Logger::ApacheLogger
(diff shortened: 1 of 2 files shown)


barberj:
@jnunemaker thoughts?

lib/httparty/logger/logger.rb
```diff
@@ -3,16 +3,23 @@
 
 module HTTParty
   module Logger
+    def self.formatters
```
jnunemaker: Thoughts on calling this loggers instead of formatters?

lib/httparty/logger/logger.rb
```diff
@@ -3,16 +3,23 @@
...
+      @formatters ||= {
+        :curl => Logger::CurlLogger,
+        :apache => Logger::ApacheLogger
+      }
+    end
+
+    def self.add_formatters custom_formatters
```
jnunemaker: I'd probably singularize this `add_logger` or `add_formater` and only add one at a time. What should the behavior be if you provide a key that already exists? This seems to just allow overriding which could be confusing. I'm wondering if we should raise if the logger already is defined.

lib/httparty/logger/logger.rb
jnunemaker: Actually, I'm ok with formatters. I as I look over how it works I see that these take a logger and format requests and responses so this works.

barberj:
@jnunemaker hows this?

I almost feel `CurlLogger` and `ApacheLogger` should be renamed to `CurlFormatter` `ApacheFormatter`

jnunemaker:
> I almost feel CurlLogger and ApacheLogger should be renamed to CurlFormatter ApacheFormatter

Agreed. I'll merge this. Feel free to whip another PR for that if you have time. If not, we'll get to it eventually.

PR #409, opened 2015-06-10
Use JSON.parse instead of JSON.load

robertclancy:
JSON.load is unsafe to use on untrusted user input:
https://www.ruby-lang.org/en/news/2013/02/22/json-dos-cve-2013-0269/
http://ruby-doc.org/stdlib-2.0.0/libdoc/json/rdoc/JSON.html#method-i-load
JSON.parse should be used instead.

This change enables quirks_mode so as to reduce breakages from this
change. Quirks mode enables the parsing of json values other than
objects or arrays, which matches the behaviour of JSON.load.

Note: there still may be breaking changes due to the differences between
load and parse:
- nil and empty string are not parsed by parse, they are parsed as nil
  by load
- NaN is not parsed by parse, it is parsed as NaN by load
- parse does not allow nesting of more than 100 to prevent stack level
  too deep errors, load has no such limit

### DIFF
--- a/lib/httparty/parser.rb
+++ b/lib/httparty/parser.rb
@@ -113,7 +113,7 @@ module HTTParty
     end
 
     def json
-      JSON.parse(body, :quirks_mode => true)
+      JSON.parse(body, :quirks_mode => true, :allow_nan => true)
     end
 
     def csv
--- a/lib/httparty/parser.rb
+++ b/lib/httparty/parser.rb
@@ -113,7 +113,7 @@ module HTTParty
     end
 
     def json
-      JSON.load(body, nil)
+      JSON.parse(body, :quirks_mode => true)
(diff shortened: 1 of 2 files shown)


jnunemaker:
We could supply a backwards compatible parser and instructions on how to use it.

robertclancy:
I had missed the fact that the nil and empty body cases are handled for all parsers:
https://github.com/jnunemaker/httparty/blob/f8cbf77cd1122ed0c97fa92081555ac7155948e7/lib/httparty/parser.rb#L101
So with the addition of `:allow_nan => true` this is backwards compatible with the old parser, bar:
- the limit on nesting
- the instantiation of ruby classes
which are the features that make `JSON.load` unsafe for untrusted inputs.

PR #413, opened 2015-06-15
Rename the loggers to formatters

barberj:
the loggers are really formatters. Rename as such


barberj:
Any idea when you will build a new gem?

PR #419, opened 2015-07-08
Always maintain method for 307 & 308 responses

### LINKED ISSUE #418: HTTP status codes 307/308 should maintain method
In the case of 307/308 statuses, method should be maintained regardless of the `:maintain_method_across_redirects` option.

The Ruby `Net` classes for these are:

- 307 - [`Net::HTTPTemporaryRedirect`](http://ruby-doc.org/stdlib/libdoc/net/http/rdoc/Net/HTTPTemporaryRedirect.html)
- 308 - [`Net::HTTPPermanentRedirect`](http://ruby-doc.org/stdlib/libdoc/net/http/rdoc/Net/HTTPPermanentRedirect.html)

307 and 308 (as proposed) parallel the behaviours of 302 and 301, but do not allow the HTTP method to change. As per [RFC 7538](https://tools.ietf.org/html/rfc7538#section-3) and [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.4.7).

billinghamj:
Resolves #418

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -276,7 +276,7 @@ module HTTParty
           unless options[:maintain_method_across_redirects] && options[:resend_on_redirect]
             self.http_method = Net::HTTP::Get
           end
-        else
+        elsif last_response.code != 307 && last_response.code != 308
           unless options[:maintain_method_across_redirects]
             self.http_method = Net::HTTP::Get
           end
@@ -305,12 +305,9 @@ module HTTParty
 
     def response_redirects?
       case last_response
-      when Net::HTTPMultipleChoice, # 300
-           Net::HTTPMovedPermanently, # 301
-           Net::HTTPFound, # 302
-           Net::HTTPSeeOther, # 303
-           Net::HTTPUseProxy, # 305
-           Net::HTTPTemporaryRedirect
+      when Net::HTTPNotModified # 304
+        false
+      when Net::HTTPRedirection
         options[:follow_redirects] && last_response.key?('location')
       end
     end
--- a/spec/httparty/request_spec.rb
+++ b/spec/httparty/request_spec.rb
@@ -528,122 +528,124 @@ RSpec.describe HTTParty::Request do
     expect(@request.perform.parsed_response).to eq('Content for you')
   end
 
-  describe "a request that 302 redirects" do
-    before(:each) do
-      @redirect = stub_response("", 302)
-      @redirect['location'] = '/foo'
-
-      @ok = stub_response('<hash><foo>bar</foo></hash>', 200)
-    end
-
-    describe "once" do
+  [300, 301, 302, 305].each do |code|
+    describe "a request that #{code} redirects" do
       before(:each) do
-        allow(@http).to receive(:request).and_return(@redirect, @ok)
+        @redirect = stub_response("", code)
+        @redirect['location'] = '/foo'
+
+        @ok = stub_response('<hash><foo>bar</foo></hash>', 200)
       end
 
-      it "should be handled by GET transparently" do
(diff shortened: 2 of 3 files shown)


billinghamj:
Due to the 308 class being missing in earlier versions of Ruby, I'll switch it to check based on the actual error code.

lib/httparty/request.rb
```diff
@@ -276,7 +276,7 @@ def handle_response(body, &block)
           unless options[:maintain_method_across_redirects] && options[:resend_on_redirect]
             self.http_method = Net::HTTP::Get
           end
-        else
+        elsif last_response.code.to_s != '307' && last_response.code.to_s != '308'
```
jnunemaker: Why the conversion to string? Seems like that just allocates more strings both for the "307" and the code.to_s when integer would be fine.

lib/httparty/request.rb
```diff
@@ -305,12 +305,9 @@ def handle_deflation
...
       case last_response
-      when Net::HTTPMultipleChoice, # 300
-           Net::HTTPMovedPermanently, # 301
-           Net::HTTPFound, # 302
-           Net::HTTPSeeOther, # 303
-           Net::HTTPUseProxy, # 305
-           Net::HTTPTemporaryRedirect
+      when Net::HTTPNotModified # 304
+        false
+      when Net::HTTPRedirection
```
jnunemaker: Is this the super class of the individually named ones previously used or something?

spec/httparty/request_spec.rb
```diff
@@ -784,6 +784,260 @@
     end
   end
 
+  describe "a request that 307 redirects" do
```
jnunemaker: We should probably add specs for the other 3xx response types. Thoughts?

lib/httparty/request.rb
billinghamj: When I was testing, the type of `code` seemed to be inconsistent. I could try undoing that and seeing if it makes any difference.

lib/httparty/request.rb
billinghamj: Correct. Also if Ruby doesn't have a specific class for the whole status code, it will just use `HTTPRedirection`. The HTTP spec also says that any unrecognized 3xx codes should be treated as 300. So other than 304, it's correct to assume that all 3xx codes should result in a redirection,

jnunemaker: :cool:

spec/httparty/request_spec.rb
billinghamj: It would cause a lot of extra code in that tests file, but no real harm in that really... I'll add testing for the others.

lib/httparty/request.rb
jnunemaker: Maybe do to_i then instead. If they are integers it doesn't hurt anything and if they aren't, it seems more efficient than the strings. Minor, but everything adds up.

spec/httparty/request_spec.rb
jnunemaker: For those that are similar you could just wrap the specs with array iteration and interpolate the code where necessary.

```ruby
[300, 301, 302, 303].each do |code|
  # spec stuff ehre
end
```

billinghamj: :+1:

lib/httparty/request.rb
billinghamj: Actually it does run consistently assuming `code` is an integer. Must have been some other issue causing it while I was testing.

billinghamj:
Made those adaptations :+1:

billinghamj:
@jnunemaker Happy with this now?

PR #424, opened 2015-07-26
add body UTF-8 check before attempting strip

landonwilkins:
repro steps:

irb(main):001:0> require 'httparty'
=> true
irb(main):002:0> url = "https://archive.org/download/beowulf/Beowulf.jpg"
=> "https://archive.org/download/beowulf/Beowulf.jpg"
irb(main):003:0> HTTParty.get(url).parsed_response
xplosions:

/.gem/ruby/2.1.5/gems/httparty-0.13.5/lib/httparty/parser.rb:103:in `strip': invalid byte sequence in UTF-8 (ArgumentError)

PR #433, opened 2015-09-19
Avoid calling String#strip on invalid Strings

pvande:
Use case: some servers have been known to serve files with
Content-Types like “video/mp4; charset=utf-8”, which HTTParty dutifully
tells ruby is a UTF-8 string.  However, before the Parser can decide
that MP4 is not a format it is capable of parsing, the response body is
tested for emptiness … after whitespace is stripped.

Since the String isn’t actually a valid UTF-8 String, this raises an
ArgumentError.  This patch fixes that error by avoiding the call on
Strings that are not validly encoded.


pvande:
This implements the same change as #424, with an appropriate test.

PR #435, opened 2015-09-21
Stop trying to deflate responses that redirect.

pvande:
Some servers (including YouTube) will send a redirect response with a
Content-Encoding header, but no body.  Since inflating an empty body
raises an exception, we should minimally work around this problem for
non-terminal redirect responses.

Note that this will still attempt to inflate a redirected response body
if HTTParty is not configured to follow redirects (which will raise
exceptions if the body is empty), and this does nothing to protect
against otherwise empty responses with a compressed Content-Encoding
(which may indicate a mis-configured server).

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -298,6 +298,8 @@ module HTTParty
 
     # Inspired by Ruby 1.9
     def handle_deflation
+      return if response_redirects?
+
       case last_response["content-encoding"]
       when "gzip", "x-gzip"
         body_io = StringIO.new(last_response.body)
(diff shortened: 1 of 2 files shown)


jnunemaker:
Why only handle redirects? Perhaps we should just avoid the gzip stuff if body is nil/empty? What was the specific exception that was raised?

pvande:
This is closely related to [this bug](https://bugs.ruby-lang.org/issues/11058) in Ruby core, particularly the "second problem" (`Zlib::BufError`).

Skipping all empty bodies is probably more correct, but I'm still new enough to the project that I wasn't 100% comfortable making that change without being aware of all the potential edge-cases (e.g. streaming bodies).

PR #436, opened 2015-09-22
Remove high-byte emoji from post-install message

007:
Since @jnunemaker will never remove the message (even though he never sees it because he runs with `--quiet`), at least make it so that the regular output is ASCII.


jnunemaker:
I don't run with --quiet. I just recommended that for the few people who are bothered by this. Is the emoji causing you issues?

jnunemaker:
I see your explanation in the comment. I have no interesting in messing up peoples installs so I'll happily merge this.

PR #438, opened 2015-10-17
Runkeeper Example

joaquinalcerro:
Added Runkeeper API Example


jnunemaker:
@joaquinalcerro thanks for the PR. There are a variety of examples, but I've tried to keep them to something that is different about httparty so they aren't just the same example for different websites. This example works similar to some of the other examples and I don't really want more to maintain so I'm going to close this. I do appreciate the PR and for anyone interested in runkeeper, they now have this PR as an example.

joaquinalcerro:
Thanks for your replay... excellent gem. Best regards.

Issue #266, opened 2014-01-31
Avoid :verify option to get overridden


guapolo:
Hi
`attach_ssl_certificates` overrides the `:verify` option value in this line https://github.com/jnunemaker/httparty/blob/master/lib/httparty/connection_adapter.rb#L147

IMHO this should not be the case, using `:pem` or `pfx`, this options should honor the previously set `:verify_mode` value.

I have a use case where using `:pem` certificates and `:verify` set to `false`   (in my case, I'm testing my code against a web service which requieres pem certificates for authentication, but the server doesn't have a signed certificate in it's testing environment).

In other words, here's the code I was using, and expected it to work:

```ruby
class MyClass
  include HTTParty
  default_params(verify: false, pem: a_certificate_and_key)
 # ...
 post(url, ...) # when posting, verify_mode is set to OpenSSL::SSL::VERIFY_PEER
end
```

but instead I need to use this declaration for it to work `OpenSSL::SSL::VERIFY_PEER = OpenSSL::SSL::VERIFY_NONE`, which is pretty ugly.

I'll submit a pull request if you agree on this behavior.

Thanks,

kurenn:
I just placed the pull request https://github.com/jnunemaker/httparty/pull/441

Had some issues with this, when working with a client, it is working now!

kurenn:
Solution has been merged with https://github.com/jnunemaker/httparty/pull/441

@jnunemaker I think you can close this issue

PR #441, opened 2015-10-30
Solves #266 - Honors the verify option to avoid overrides

kurenn:
####What does this PR do?

* Honors the verify option to prevent `http.verify_mode` when using `pem` or `pk12` options


jnunemaker:
Looks like there are failures on travis ci now. If we could get those fixed up, I'd be happy to merge this.

kurenn:
I'll update the code =)

jnunemaker:
Won't this break backwords compatibility for those using verify_peer for this? I think by tests I was thinking that verify works too, but verify_peer is used first if set to ensure it doesn't break anything. Or am I misunderstanding? Honestly I didn't write most of this code, I just want to make sure I don't break things for people.

kurenn:
Ok, you are right! I'll change the code and push it!

PR #445, opened 2015-11-12
added is_a? to the response class and DRYed up the specs a bit

will3216:
+ Moved the other instance method specs to more appropriate locations based on context
+ DRY'd the specs from redundant instantiation calls
+ Added an is_a? instance method to the Response class


will3216:
Lol, probably should have checked the PR's first, deleting as duplicate.

PR #449, opened 2015-11-21
Adds docs directory with SSL configuration help

kurenn:
#### What does this PR do?

* Adds `docs` directory
* Adds `README.md` file under `docs` with documentation about SSL


kurenn:
Can't reproduce this, it breaks on ruby 1.9.3p551, tried on local and it works! Any suggestion?

PR #450, opened 2015-11-30
Don't send basic auth header when redirecting to different hosts

### LINKED ISSUE #404: HTTParty leaks basic auth header to redirect target
Calling `get()` with the `basic_auth` option set will automatically follow redirects and send the basic auth header to both the first host and the redirect target, even if on different domains. This is a security risk and should either be disabled or at least be configurable.

linrock:
Fixes #404 . This behavior is consistent with cURL:
https://github.com/bagder/curl/blob/6beb0eee/lib/http.c#L710

Another way to demonstrate the new behavior:
```python
import SimpleHTTPServer

class RequestHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):

    def do_GET(self):
        if self.path == "/":
            self.send_response(302)
            self.send_header("Location",   "http://127.0.0.1:8000/2")
            print self.headers
        elif self.path == "/2":
            self.send_response(302)
            self.send_header("Location",   "http://localhost:8000/3")
            print self.headers
        else:
            self.send_response(200)
            print self.headers

if __name__ == '__main__':
    SimpleHTTPServer.test(HandlerClass=RequestHandler)
```
Send a request that uses basic auth:
```ruby
HTTParty.get "http://username:password@127.0.0.1:8000"
```
Server responses:
```http
# First request sends the Authorization header
GET / HTTP/1.1 302
Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
Accept: */*
User-Agent: Ruby
Authorization: Basic dXNlcm5hbWU6cGFzc3dvcmQ=
Connection: close
Host: 127.0.0.1:8000

# Second request also sends the auth header since it redirects to the same host 
GET /2 HTTP/1.1 302
Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
Accept: */*
User-Agent: Ruby
Authorization: Basic dXNlcm5hbWU6cGFzc3dvcmQ=
Connection: close
Host: 127.0.0.1:8000

# Third request doesn't send the auth header because the host changed
GET /3 HTTP/1.1 200
Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
Accept: */*
User-Agent: Ruby
Connection: close
Host: localhost:8000
```

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -128,6 +128,7 @@ module HTTParty
       end
 
       handle_deflation unless http_method == Net::HTTP::Head
+      handle_host_redirection if response_redirects?
       handle_response(chunked_body, &block)
     end
 
@@ -174,7 +175,7 @@ module HTTParty
       @raw_request.body = body if body
       @raw_request.body_stream = options[:body_stream] if options[:body_stream]
       @raw_request.initialize_http_header(options[:headers].to_hash) if options[:headers].respond_to?(:to_hash)
-      @raw_request.basic_auth(username, password) if options[:basic_auth]
+      @raw_request.basic_auth(username, password) if options[:basic_auth] && send_authorization_header?
       setup_digest_auth if options[:digest_auth]
     end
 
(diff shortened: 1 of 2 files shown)


jnunemaker:
Thanks!

PR #452, opened 2015-12-01
Version 1.0.0

henrik:
This library is 7 1/2 years old – might it be time to consider it mature? :)

This is not intended as a dig but as a practical concern: I noticed the pre-1.0 version when updating gems and reflected on the fact that if it was post 1.0 and used semantic versioning, I could have more easily judged the impact of the upgrade.

I know the contribution guidelines say don't mess with the version, but it's hard to avoid in this case :D

### DIFF
--- a/lib/httparty/version.rb
+++ b/lib/httparty/version.rb
@@ -1,3 +1,3 @@
 module HTTParty
-  VERSION = "0.13.7"
+  VERSION = "1.0.0"
 end


jnunemaker:
Fair. I've thought about it for a while, but it feels so permanent and big. The hilarious thing is I'm not sure I even have a 1.0 gem. Haha. I'll think about this for a bit and see if there is anything big that should be changed prior.

jnunemaker:
By bit I mean I've added a task to revisit in a week.

henrik:
:) :thumbsup:

jnunemaker:
Haven't had time to review this yet, but still on my list.

jnunemaker:
I can't really think of a reason not to do this. I'll try to go to 1.0.0 for the next release.

nj:
Time flies when having fun :grin:

jnunemaker:
I'll try to go to 1.0 soon. Closing this though since it is a minor change.

henrik:
@jnunemaker 🎉 Godspeed!

Issue #454, opened 2015-12-21
Feature Proposal: Raise on some HTTP status codes


dccunha:
I would like if HTTParty throw an exception on some HTTP status codes. I thought in an option that would turn this feature on and define what status codes would throw exceptions.

Something like this:

```ruby
class Foo
  include HTTParty
  raises_on status_codes: (400..599).to_a
end
```

What do you think of this? Can I send a pull request following this idea?

jnunemaker:
I'd be open to seeing a PR, but I'm not really sure I like this idea. I'd probably rather just see a layer on top at the application level or something that does this if you want it.

jnunemaker:
Closing the issue, but thanks for the proposal and do let me know if you whip something up.

PR #455, opened 2015-12-29
Raise on some HTTP status codes

dccunha:
#### What?

Implements the proposal #454   

#### Why?

When you use httparty in many different projects it starts to become tedious to add a verification on every request. To accomplish this, you have to do something like this: 

```ruby
module Api
  class Base
    include HTTParty

    def get(path, options = {})
      raises_unless_success { self.class.get path, options }
    end

    private

     def raises_unless_success
      httparty = yield
      fail HTTParty::Error, "Code #{httparty.code} - #{httparty.body}" unless
        httparty.response.is_a?(Net::HTTPSuccess)
      httparty
    end

  end
end
```

I believe this is not a straight solution.

### My idea

```ruby
module Api
  class Base
    include HTTParty
    raise_on (400..599).to_a

    def get(path, options = {})
      self.class.get path, options
    end
  end
end
```

### DIFF
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -72,6 +72,16 @@ module HTTParty
       default_options[:log_format] = format
     end
 
+    # Raises HTTParty::ResponseError if response's code matches this statuses
+    #
+    #   class Foo
+    #     include HTTParty
+    #     raise_on [404, 500]
+    #   end
+    def raise_on(codes = [])
+      default_options[:raise_on] = *codes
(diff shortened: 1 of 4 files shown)


jnunemaker:
I actually don't like automatic raising of exceptions and prefer explicitly checking errors in each case, but I have a feeling that many users of httparty will like this and it is optional, so I'm going to merge. Thanks for the PR!

PR #457, opened 2016-01-13
Allow nil http_proxyaddr to pass through to Net::HTTP

pic:
This is useful when a proxy is used by default but you do not want to use it for some requests


lib/httparty/connection_adapter.rb
```diff
@@ -76,7 +76,7 @@ def initialize(uri, options = {})
     def connection
       host = clean_host(uri.host)
       port = uri.port || (uri.scheme == 'https' ? 443 : 80)
-      if options[:http_proxyaddr]
+      if options.has_key?(:http_proxyaddr)
```
jnunemaker: I'm going to change this to just key? but I'll pull and fix. Thanks!

Issue #459, opened 2016-01-29
to_cookie_string mutates CookieHash values


sethwoodworth:
https://github.com/jnunemaker/httparty/blob/6f4d6ea4a2707a4f1466f45bf5c67556cdbed2b7/lib/httparty/cookie_hash.rb#L19

```ruby
[20] pry(main)> h
=> {:session=>"91e25e8b-6e32-418d-c72f-2d18adf041cd", :"Max-Age"=>"15552000", :cart=>"91e25e8b-6e32-418d-c72f-2d18adf041cd", :httponly=>nil, :Path=>"/", :secure=>nil}
[21] pry(main)> h.to_cookie_string
=> "session=91e25e8b-6e32-418d-c72f-2d18adf041cd; Max-Age=15552000; cart=91e25e8b-6e32-418d-c72f-2d18adf041cd"
[22] pry(main)> h
=> {:session=>"91e25e8b-6e32-418d-c72f-2d18adf041cd", :"Max-Age"=>"15552000", :cart=>"91e25e8b-6e32-418d-c72f-2d18adf041cd"}
```

I would not have expected `to_cookie_string` to be removing keys from a CookieHash.  I understand removing keys that are not in `CLIENT_COOKIES` in the output of the method, but not mutating the parent object.

jnunemaker:
That is really odd. I'm open to a PR that fixes it. Thanks for filing.

mikeastock:
I think this has been resolved by #460

PR #460, opened 2016-01-30
Use reject instead of delete_if to prevent mutation

mikeastock:
Solves https://github.com/jnunemaker/httparty/issues/459. `Hash#reject` returns a new hash instead of mutating. http://ruby-doc.org/core-2.2.0/Hash.html#method-i-reject


tibbon:
This looks great. yay!

PR #461, opened 2016-01-31
Add Ruby 2.3 to Travis matrix

mikeastock:


mikeastock:
Looks like rvm doesn't have a binary for Ruby 2.3. Do you prefer to only use binaries?

jnunemaker:
I noticed this on a different project (that 2.3 isn't there). I don't really care how it works. If you know of another way to make 2.3 work, I'm open to that.

mikeastock:
So it seems that if you specify the patch version `rvm` can find and install Ruby 2.3.0.

jnunemaker:
> So it seems that if you specify the patch version rvm can find and install Ruby 2.3.0.

Seems fine to me. I guess only downside is it won't upgrade with patch upgrades.

mikeastock:
This PR should be good to go then.

jnunemaker:
Thanks!

PR #464, opened 2016-02-21
Remove json gem

deepj:
JSON support is in Ruby since 1.9 so there is no need to have json gem as a dependency.

Inspired by this Rack’s PR https://github.com/rack/rack/pull/1011


jnunemaker:
Seems fine. Most people will probably already have the gem anyway. Anything to worry about here?

deepj:
@jnunemaker I don't think so. The dependency was removed in Rails as well. See: https://github.com/rails/rails/pull/23453

PR #465, opened 2016-03-11
Implemented support for MKCOL method

nj:
`MKCOL` - part of the [HTTP Extensions for Distributed Authoring](http://www.webdav.org/specs/rfc2518.html) (also known as WebDAV) - allow creation of a new collection resource at the location specified.

By extending with the `MKCOL` method, will cover (most) basic WebDAV implementation and make `httparty` more generic and increase the fun :smile:

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -9,7 +9,8 @@ module HTTParty
       Net::HTTP::Head,
       Net::HTTP::Options,
       Net::HTTP::Move,
-      Net::HTTP::Copy
+      Net::HTTP::Copy,
+      Net::HTTP::Mkcol
     ]
 
     SupportedURISchemes  = ['http', 'https', 'webcal', nil]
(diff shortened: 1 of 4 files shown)


lib/httparty/request.rb
```diff
@@ -9,7 +9,8 @@ class Request #:nodoc:
       Net::HTTP::Head,
       Net::HTTP::Options,
       Net::HTTP::Move,
-      Net::HTTP::Copy
+      Net::HTTP::Copy,
+      Net::HTTP::Mkcol
```
jnunemaker: Can we add a trailing comma here so we don't end up with a diff like this next time something is added?

jnunemaker:
Seems fine. I know nothing about it, so I guess we'll see how it goes with maintenance on it going forward.

lib/httparty/request.rb
nj: Had actually thought about it, but decided to avoid in the first place to keep konsistency with the original.

It has been added now in 5f94172 :smiley:

jnunemaker:
Thanks!

PR #466, opened 2016-03-29
Fix base uri options not normalized

jgoyon:
Calling `HTTParty::ClassMethods.base_uri` or passing `base_uri` to a request method do not behave the same way. The class method call normalize the uri, while passing options does not. 

This will throw an error due to the url no being valid (it lacks the scheme in the url)

```ruby
HTTParty.get('/resource', {
  base_uri: 'foo.com',
  body: {}
})
```

But this will works fine,[`HTTParty.normalize_base_uri`](https://github.com/jnunemaker/httparty/blob/master/lib/httparty.rb#L107) takes care of adding the scheme  
```ruby
class Myclass
  include HTTParty
  base_uri 'foo.com'

  def get_it
    get('/resource', {
      body: {}
    })
  end
end
```

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -94,7 +94,7 @@ module HTTParty
         base_uri += ":#{@last_uri.port}" if @last_uri.port != 80
         base_uri
       else
-        options[:base_uri]
+        options[:base_uri] && HTTParty.normalize_base_uri(options[:base_uri])
       end
     end
 
--- a/spec/httparty/request_spec.rb
+++ b/spec/httparty/request_spec.rb
@@ -168,6 +168,14 @@ RSpec.describe HTTParty::Request do
       request.send(:setup_raw_request)
       expect(request.instance_variable_get(:@raw_request).body_stream).to eq(stream)
     end
+
+    it 'should normalize base uri when specified as request option' do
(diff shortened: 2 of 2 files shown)


jnunemaker:
I never thought about the fact that base_uri would even work when used that way. It was only ever intended to be called from class method as far as I can remember. Seems fine to merge this though. The only downside is double normalizing something that has already been normalized, but that probably isn't a big deal.

jgoyon:
Thanks for merging.

To give some context, I have many endpoints that implement the same API and a ruby wrapper around that API having more or less that structure:

```ruby
class MyAPI

  include HTTParty

  def initialize(base_uri)
    @base_uri = base_uri
  end

  def token
    #logic to get token
  end

  def resource_a(params)
    query = params.merge!({
      token: token
    })
    options = {
      base_uri: @base_uri,
      query: query
    }
    resp = self.class.get('/resource_a', options)
    resp['data']
  end

end
```
As I have many possible uri, I do not see an easy/clean way to use the `base_uri` class method option. So I need to pass it as a parameter to `MyAPI` on initialisation. But I found that keeping the `base_uri` as part as the `options` hash made the code more readable.

jnunemaker:
@jgoyon yeah for that I usually just do a full get, ie: self.class.get("http://#{@base_uri}/resource_a", ...). Seems fine to use the option I guess, I just never planned on anyone doing it that way.

PR #474, opened 2016-05-17
Update README.md

mmplisskin:
It is not clear that the proper case for requiring this module is as follows.


jnunemaker:
This is not correct. The correct usage is `require "httparty"`. It should be lower case.

PR #477, opened 2016-06-08
Allow empty array to be used as a params in get requests.

### LINKED ISSUE #469: Empty array is ignored when passed as a param
```
res = HTTParty.get('http://google.com', query: {fake_array: []})
res.request.last_uri.to_s == 'http://www.google.com/'
```
but...
```
res = HTTParty.get('http://google.com', query: {fake_array: [1]})
res.request.last_uri.to_s == 'http://www.google.com/?fake_array%5B%5D=1'
```

Is this intended behavior? Is this a limitation of HTTP?

dmitryrck:
Fixes #469 

Now we have:

```
>> res = HTTParty.get('http://github.com', query: {fake_array: []})
# […]
>> res.request.last_uri.to_s
=> "https://github.com/?fake_array[]="
```

And:

```
>> res = HTTParty.get('http://github.com', query: {fake_array: [1]})
# […]
>> res.request.last_uri.to_s
=> "https://github.com/?fake_array[]=1"
```

Besides this fix I'd like to know what happens if you have something like this: `HTTParty::Request.new(Net::HTTP::Get, 'http://google.com', query: { fake_array: [""] })`

### DIFF
--- a/lib/httparty/hash_conversions.rb
+++ b/lib/httparty/hash_conversions.rb
@@ -26,7 +26,11 @@ module HTTParty
       stack = []
 
       if value.respond_to?(:to_ary)
-        param << value.to_ary.map { |element| normalize_param("#{key}[]", element) }.join
+        param << if value.empty?
+                   "#{key}[]=&"
+                 else
+                   value.to_ary.map { |element| normalize_param("#{key}[]", element) }.join
+                 end
       elsif value.respond_to?(:to_hash)
         stack << [key, value.to_hash]
       else
--- a/spec/httparty/hash_conversions_spec.rb
+++ b/spec/httparty/hash_conversions_spec.rb
@@ -22,6 +22,14 @@ RSpec.describe HTTParty::HashConversions do
       end
     end
 
+    context "value is an empty array" do
+      it "creates a params string" do
+        expect(
+          HTTParty::HashConversions.normalize_param(:people, [])
+        ).to eq("people[]=&")
+      end
+    end
+
     context "value is hash" do
       it "creates a params string" do
         expect(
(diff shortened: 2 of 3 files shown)


lib/httparty/request.rb
```diff
@@ -78,7 +78,11 @@ def uri
 
       # avoid double query string on redirects [#12]
       unless redirect
-        new_uri.query = query_string(new_uri)
+        possible_query_string = query_string(new_uri)
+
+        if possible_query_string != nil && possible_query_string != ""
```
jnunemaker: Out of curiousity, what specifically made it so you need this?

dmitryrck: Well… I was doing baby steps and the change was required, but I double-checked and it is not required anymore :)

lib/httparty/request.rb
```diff
@@ -201,7 +201,9 @@ def query_string(uri)
         query_string_parts << options[:query] unless options[:query].nil?
       end
 
-      query_string_parts.reject!(&:empty?) unless query_string_parts == [""]
+      query_string_parts.reject! { |part|
```
jnunemaker: I checked locally and this isn't needed to make any specs pass. Do we need a new spec or is this extra as well?

dmitryrck: Weird… I remember I needed this :/

Let me check :)

dmitryrck: Hey @jnunemaker thank you for the great review, it was necessary to change only `HTTParty::HashConversions` eventually :)

DaftTrick:
In this PR you mention that change of behavior should be only for get method.
But this PR also changed working with POST request also. Is that designed that way or just a side-bug?

dmitryrck:
@DaftTrick I did all the tests based on GET, but I did not remember of seeing anything to mean that the behavior should be only to GET, so I think this code should work to POST too.

Why? Any problem with that?

PR #482, opened 2016-06-21
Add status predicate methods to Response#respond_to?

akamike:
Expands the existing `respond_to?` first-party method list to include the HTTP status predicates. Useful for things like RSpec, which check `respond_to?` when using predicate matchers - beyond that it makes for a more consistent and predictable object (If I can call it, it should `respond_to?`).

Because `Response` inherits from `BasicObject` we miss out on quite a few of these pre-made tools from `Object`, though it looked like moving away from `BasicObject` was a bigger task and would require a bit more debate. I figured this was the simplest change to make.

I have not checked for other methods that could be added to `RESPOND_TO_METHODS`, my concern was with the predicate methods.

### DIFF
--- a/spec/httparty/response_spec.rb
+++ b/spec/httparty/response_spec.rb
@@ -107,6 +107,11 @@ RSpec.describe HTTParty::Response do
     expect(response.respond_to?(:parsed_response)).to be_truthy
   end
 
+  it "responds to predicates" do
+    response = HTTParty::Response.new(@request_object, @response_object, @parsed_response)
+    expect(response.respond_to?(:success?)).to be_truthy
+  end
+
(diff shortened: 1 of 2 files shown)


spec/httparty/response_spec.rb
```diff
@@ -107,6 +107,11 @@
     expect(response.respond_to?(:parsed_response)).to be_truthy
   end
 
+  it "responds to predicates" do
+    response = HTTParty::Response.new(@request_object, @response_object, @parsed_response)
+    expect(response.respond_to?(:success?)).to be_truthy
```
jnunemaker: Kind of concerned that we'll unknowingly break compatibility since only success? is tested, but I guess we can roll and see what happens.

PR #485, opened 2016-07-12
Added example for stream mode

fanantoxa:
It was a bit hard to understand how to work with stream mode so I've added example for stream mode.

### DIFF
--- a/examples/stream_download.rb
+++ b/examples/stream_download.rb
@@ -1,0 +1,24 @@
+dir = File.expand_path(File.join(File.dirname(__FILE__), '..', 'lib'))
+require File.join(dir, 'httparty')
+require 'pp'
+
+class StreamDownload
+  include HTTParty
+
+  base_uri 'https://cdn.kernel.org/pub/linux/kernel/v4.x'
+end
+
+# download file linux-4.6.4.tar.xz without using the memory
+response = nil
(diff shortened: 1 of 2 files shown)


fanantoxa:
@jnunemaker, all tests except on ruby 1.9.3 pass.
Cloud you please merge example to master?
Also added issue here: https://github.com/jnunemaker/httparty/issues/486

examples/stream_download.rb
```diff
@@ -0,0 +1,23 @@
+dir = File.expand_path(File.join(File.dirname(__FILE__), '..', 'lib'))
+require File.join(dir, 'httparty')
+require 'pp'
+
+class StreamDownload
+  include HTTParty
+  stream_body = true
```
jnunemaker: I don't believe this is correct. [stream_body is an option for requests](https://github.com/jnunemaker/httparty/blob/0d808c7f3a68b8d091506e22003fc00dd84bf6a6/spec/httparty_spec.rb#L777-L788), not a class level method. Also, class level methods don't use ='s as that would mean assignment, not a method call.

fanantoxa:
@jnunemaker I've updated code with your additions

fanantoxa:
@jnunemaker please take a look when you'll find time please )

PR #487, opened 2016-07-28
Require ERB where it's used

pclalv:
HTTParty users that only require parts of it, e.g. `HashConversions` via `require 'httparty/hash_conversions'`, will see `uninitialized constant HTTParty::HashConversions::ERB` if their code does not itself `require 'erb'`. this is because ERB is only properly required when HTTParty is required with `require 'httparty'`. this PR allows users to use the parts of HTTParty that require ERB without having to `require 'erb'` in their code.

### DIFF
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -6,7 +6,6 @@ require 'zlib'
 require 'multi_xml'
 require 'json'
 require 'csv'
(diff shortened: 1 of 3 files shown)


jnunemaker:
Thanks!

PR #492, opened 2016-08-24
Deal with network-path reference redirects

### LINKED ISSUE #491: Redirection breaks when location header is a network-path reference
Whenever the location header in a redirect is a network-path reference (something like "//www.example.com"), a RedirectionTooDeep error occurs. The reason for this is "//www.example.com" is considered a [relative URI](https://tools.ietf.org/html/rfc3986#section-4.2). This causes [`HTTParty::Request#uri`](https://github.com/jnunemaker/httparty/blob/6049c5e798277de0775b286553a2802ec87bd883/lib/httparty/request.rb#L79) to append it to the end of the existing URI. This results in a series of redirects as follows:

http://example.com
http://example.com//www.example.com
http://example.com//www.example.com//www.example.com
etc.
(the location header for all of these is "//www.example.com")

For an in-the-wild example, I encountered this with `HTTParty.get("http://zillow.com")`

capicue:
fixes #491

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -76,7 +76,13 @@ module HTTParty
         path.path = last_uri_host + path.path
       end
 
-      new_uri = path.relative? ? options[:uri_adapter].parse("#{base_uri}#{path}") : path.clone
+      if path.relative? && path.host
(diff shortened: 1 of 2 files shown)


jnunemaker:
:bow:

PR #493, opened 2016-08-31
Docs fix: streaming example doesn't use PORO class

vergenzt:


vergenzt:
(Looks like there was a missing newline at end of file too.)

jnunemaker:
:bow:

PR #499, opened 2016-10-26
Path options for TLS cert added

kvokka:
When I've tried to use https://github.com/NARKOZ/gitlab gem, I needed to serialize TLS options to global variable. It is very ugly, so, I added `_path` option for `p12` and `pem` for key and password files, so, user may store it in options directly, or use path. Syntax will be like `{ pem_path: '/path/to/cert'}`

I hope you'll like it

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -98,3 +98,13 @@ class Client
 	end
 end
 ```
+
+### Using with `gitlab` gem
+
+Add this to `~/.bashrc`, `_path` preffix avaiable for `p12` and `pem` only. If you need home path, use `ENV['HOME']` instead of `~`.
+
+```
+
+export GITLAB_API_HTTPARTY_OPTIONS="{verify: true, p12_path: '/path/to/key', p12_password_path: '/path/to/password' }"
+
+```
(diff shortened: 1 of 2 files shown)


kvokka:
gitlab gem will support this, if you only merge PR, or, say please, what you dislike please. Thank you

lib/httparty/connection_adapter.rb
```diff
@@ -195,5 +197,11 @@ def attach_ssl_certificates(http, options)
         end
       end
     end
+
+    def normalize_key_files options
```
jnunemaker: I don't understand what this is doing. Is it changing certificate options by reading files somewhere on the filesystem? 

Could you perhaps add a test or two to show how this works and ensure that I don't break it in the future. These options are not something I use regularly, so I want to make sure I don't mess it up going forward.

docs/README.md
```diff
@@ -97,4 +97,14 @@ class Client
 		# get("resources", verify_peer: false)
 	end
 end
-```
\ No newline at end of file
+```
+
+### Using with `gitlab` gem
```
jnunemaker: I'm not interested in something specific to another gem living in the docs here. I think it would be better to have this in the gitlab gem docs or am I misunderstanding something?

PR #501, opened 2016-11-08
add spacing to example hashes

kholbekj:
Did this because the nested hashes with inconsistent styles was driving me mad. 

Also willing to create the reverse version as PR if deemed better, eg. always having no space between hash opening and closing brackets:

`{herp: derp}`

Just, please, one style 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ class StackExchange
   base_uri 'api.stackexchange.com'
 
   def initialize(service, page)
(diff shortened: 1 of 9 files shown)


jnunemaker:
I usually prefer no spaces, but I'm not very picky about this stuff anymore. Merging to avoid any more work for you (like reversing). :) Thanks for the PR!

Issue #500, opened 2016-11-03
How to symbolize JSON keys or suppress parsing?


georgeu2000:
I would like JSON response keys to be symbols, but did not see a way of doing that.

Alternately, how to suppress parsing JSON?

    HTTParty.get 'http://example.com'
    => { "foo" => "bar" }

jnunemaker:
> I would like JSON response keys to be symbols, but did not see a way of doing that.

There is no automatic way to do key conversion to symbols and I don't plan on adding it. Newer versions of ruby have symbol gc, but older ones don't which would make your application open to attack simply by an external service returning lots of different keys, which would increase memory usage.

If you are using from Rails or have access to ActiveSupport, you can use `symbolize_keys` to convert keys.

> Alternately, how to suppress parsing JSON?

The format is inferred from the content type header in the response. You can see a mapping of supported formats [here](https://github.com/jnunemaker/httparty/blob/247d541f93e27d96e8226ec222a0c488886d1017/lib/httparty/parser.rb#L40-L52). If you want to override that, you can do so by passing format as plain:

```ruby
HTTParty.get 'http://example.com', format: :plain
```

georgeu2000:
OK, thanks. Overriding with `format: :plain` is great. 

I don't see anything in the docs about automatic parsing or overriding. Would you be open to a PR?

jnunemaker:
Always open to new docs if they make sense and will help!

PR #503, opened 2016-11-11
Update README with parsing JSON keys as symbols

georgeu2000:
How to symbolize JSON keys or suppress parsing?

https://github.com/jnunemaker/httparty/issues/500

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,14 +47,6 @@ puts stack_exchange.users
 
 See the [examples directory](http://github.com/jnunemaker/httparty/tree/master/examples) for even more goodies.
 
-## JSON Parsing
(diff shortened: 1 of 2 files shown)


jnunemaker:
Thank for the PR. Really appreciate you taking the time to make httparty better for the next person. 

That said, I'm not sure I want to mention something so specific like this in the primary readme, as it has rarely come up. Perhaps it would be better in the [docs readme](https://github.com/jnunemaker/httparty/blob/master/docs/README.md)? What do you think?

georgeu2000:
I'm not really sure what the purpose of limiting the primary readme, when it is already too short. Also, the docs readme is only about SSL.

Having said that, I'll be happy to move it.

georgeu2000:
@jnunemaker - I ran into this again today. Is it OK to merge?

Issue #506, opened 2017-01-05
HTTParty permanently overrides default request headers of Net::HTTP


qsona:
## Version
httparty 0.13.7
(Ruby on Rails 4.2.6)

## Reproduce Steps

1) `HTTParty.get('http://httpstat.us/204', debug_output: $stdout)`
request log:
`<- "GET /204 HTTP/1.1\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nAccept: */*\r\nUser-Agent: Ruby\r\nConnection: close\r\nHost: httpstat.us\r\n\r\n"`

2) `HTTParty.get('http://httpstat.us/204', headers: {}, debug_output: $stdout)`
request log:
`<- "GET /204 HTTP/1.1\r\nConnection: close\r\nHost: httpstat.us\r\n\r\n"`

3) `HTTParty.get('http://httpstat.us/204', debug_output: $stdout)`
request log:
`<- "GET /204 HTTP/1.1\r\nConnection: close\r\nHost: httpstat.us\r\n\r\n"`

On the second request, I specified `headers: {}` option, then the default headers (Accept-Encoding, Accept, User-Agent) are removed. And after that, even if I don't specified headers option, these headers don't come back.

jnunemaker:
Closed by https://github.com/jnunemaker/httparty/pull/518

PR #508, opened 2017-01-18
Reset default request headers

tomatoturnip:
Made a fix to reset default request headers re: #506

### DIFF
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -553,10 +553,16 @@ module HTTParty
     end
 
     def perform_request(http_method, path, options, &block) #:nodoc:
+      reset_default_headers
+
       options = ModuleInheritableAttributes.hash_deep_dup(default_options).merge(options)
       process_headers(options)
       process_cookies(options)
       Request.new(http_method, path, options).perform(&block)
+    end
+
+    def reset_default_headers
+      default_options.delete(:headers) if default_options[:headers] == {}
     end
 
     def process_headers(options)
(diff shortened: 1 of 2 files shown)


CJStadler:
@tloggins and I ran into this issue and I've verified that this PR fixes it for us. Thanks @tomatoturnip!

CJStadler:
@jnunemaker any chance you could review this? It would be much appreciated 🙇

jnunemaker:
Extremely sorry I overlooked this. Not sure what went wrong.

So the goal of this is to not pass headers to net/http if none are specified in httparty so that net/http's default headers are used?

CJStadler:
@jnunemaker that is correct.

Instead of doing this cleanup before every request perhaps it would be cleaner to either

1. Change `process_headers` to use `default_options[:headers]` instead of `headers`. 
2. Change `headers` to not modify `default_options[:headers]` when it does not receive an arg.

CJStadler:
@jnunemaker sorry to ping you on this again, but it caused another unexpected build failure today. I'm happy to open a PR with either of the above two solutions, if you'd prefer that to the current solution.

jnunemaker:
@CJStadler no prob! I would be fine with 1 or 2. Which feels better to you?

jnunemaker:
@tomatoturnip I'm going to close this one. Again, very sorry I overlooked it. Once I did look it over, I didn't quite understand the problem. Now that I do, I feel that @CJStadler suggestions would be better. My main problem with this one is that it deletes things, which feels a little brute force. I'd rather see some logic that avoids the delete. I'd be happy to merge this if you want to update it or I can wait for @CJStadler's PR.

tomatoturnip:
@jnunemaker No worries. @CJStadler can create the PR. His suggestions sound good. Thanks guys! :)

PR #518, opened 2017-04-12
Do not overwrite default headers unless specified

CJStadler:
Previously, calling `HTTParty.headers` without an argument set
`HTTParty.default_options[:headers]` to an empty hash. Now the
default headers will only be set to an empty hash when one is
explicitly provided, otherwise they will remain `nil`.

The previous behavior was problematic because the default headers
for a request could either be `nil` or `{}`, depending on whether
`HTTParty.headers` had been called. When they are `nil` the
default headers from Net::HTTP are used.

@jnunemaker any feedback is appreciated! This feels to me like a bit more of a workaround than a real solution, but I don't see a cleaner solution without a somewhat larger refactor.

Thanks to @tomatoturnip for the original [PR](https://github.com/jnunemaker/httparty/pull/508).

### DIFF
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -212,10 +212,14 @@ module HTTParty
     #     include HTTParty
     #     headers 'Accept' => 'text/html'
     #   end
-    def headers(h = {})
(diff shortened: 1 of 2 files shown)


CJStadler:
Thanks @jnunemaker! 🤘

PR #519, opened 2017-04-17
Add JSON_API_QUERY_STRING_NORMALIZER

Rsullivan00:
JSON API generally handles query string list parameters by comma-separating the values:

```
/articles?include=author&fields[articles]=title,body,author&fields[people]=name 
(with encoded '[' and ']' characters)
```

The default serializer and the `NON_RAILS_QUERY_STRING_NORMALIZER` currently creates duplicate query params with each value in a list:
```
Default
/articles?include=author&fields[articles][]=title&fields[articles][]=body&fields[articles][]=author&fields[people]=name 
NON_RAILS
/articles?include=author&fields[articles]=title&fields[articles]=body&fields[articles]=author&fields[people]=name 
```

This PR adds another query string normalizer that will produce URLs of the first format, which will work out of the box with JSON API-compliant APIs.

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -29,6 +29,19 @@ module HTTParty
       end.flatten.join('&')
     end
 
+    JSON_API_QUERY_STRING_NORMALIZER = proc do |query|
+      Array(query).sort_by { |a| a[0].to_s }.map do |key, value|
+        if value.nil?
+          key.to_s
+        elsif value.respond_to?(:to_ary)
+          values = value.to_ary.map{|v| ERB::Util.url_encode(v.to_s)}
+          "#{key}=#{values.join(',')}"
+        else
+          HashConversions.to_params(key => value)
+        end
+      end.flatten.join('&')
+    end
+
     attr_accessor :http_method, :options, :last_response, :redirect, :last_uri
     attr_reader :path
 
--- a/spec/httparty/request_spec.rb
+++ b/spec/httparty/request_spec.rb
@@ -33,6 +33,34 @@ RSpec.describe HTTParty::Request do
     end
   end
 
+  describe "::JSON_API_QUERY_STRING_NORMALIZER" do
+    let(:normalizer) { HTTParty::Request::JSON_API_QUERY_STRING_NORMALIZER }
+
+    it "doesn't modify strings" do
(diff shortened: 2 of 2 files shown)


jnunemaker:
First, thanks so much for taking the time to submit this. I really appreciate PRs. 

I'm wondering, is this really common? Would you mind dropping a few examples of APIs that do this? Perhaps it would be best as an example in the examples directory rather than in httparty itself since this is the first time anyone has brought it up in the many years httparty has been around?

Rsullivan00:
At my work, we rely heavily on [JSONAPI::Resources](https://github.com/cerebris/jsonapi-resources) to build our APIs.

JSONAPI::Resources does not work with the Default/NON_RAILS query strings; it requires comma-separated list values.

The JSON-API spec explicitly names certain things ([include param values](http://jsonapi.org/format/#fetching-includes), [sparse fieldset requests](http://jsonapi.org/format/#fetching-sparse-fieldsets), and [multiple sorting fields](http://jsonapi.org/format/#fetching-sorting)) that must be comma-separated. In theory, that means that any JSON-API compliant API will require a query-string in this format.

As for other APIs, I did some Googling, and it looks like Facebook's marketing API uses comma-separated values for sparse fieldsets:

```zsh
curl -G \
-d "level=ad" \
-d "fields=impressions,ad_id" \
-d "access_token=<ACCESS_TOKEN>" \
"https://graph.facebook.com/<API_VERSION>/<CAMPAIGN_ID>/insights"
```

[Spotify comma-separates filter values](https://developer.spotify.com/web-api/get-several-artists/):
```zsh
curl -X GET "https://api.spotify.com/v1/artists?ids=0oSGxfWSnnOXhD2fKuz2Gy,3dBVyJ7JuOMt4GE9607Qin"
```

I think comma-separating values is a common practice, although I haven't verified whether or not the duplicate-param approach is also supported by these APIs.

That said, I don't know whether this alternate query parser belongs in `HTTParty::Response`, or if it's better off as an example or addon-type thing.

It might make sense to move these alternate query parsers out into a module like `HTTParty::QueryParsers`.

What do you think?

jnunemaker:
Thank you so much for the context. I didn't even realize there was a spec for json api. I'm cool with merging as is.

PR #520, opened 2017-04-17
[#498] Handle utf-8 bom for json parsing

### LINKED ISSUE #498: BOM characters \xEF\xBB\xBF
When the response body contains BOM characters the Json parsing fails.
This response breaks it:

"\xEF\xBB\xBF[{\"id...

I get:

Exception Encoding::UndefinedConversionError
Error "\xEF" from ASCII-8BIT to UTF-8

I'm using httparty 0.14.0
Ruby 2.2.4

Any hints?

I can make the request with other libraries and clean up the BOM, but httparty is a dependency of other gem I use.

Rsullivan00:
Closes #498 

Currently, HTTParty errors when you try to parse JSON that has a utf-8 BOM. Since .NET applications automatically prepend the BOM characters, this severely limits the usefulness of HTTParty when talking to Windows-based JSON services.

This PR resolves that issue by stripping the BOM characters before parsing the JSON response.

### DIFF
--- a/lib/httparty/parser.rb
+++ b/lib/httparty/parser.rb
@@ -114,8 +114,10 @@ module HTTParty
       MultiXml.parse(body)
     end
 
+    UTF8_BOM = "\xEF\xBB\xBF".freeze
+
     def json
-      JSON.parse(body, :quirks_mode => true, :allow_nan => true)
+      JSON.parse(body.gsub(/^#{UTF8_BOM}/, ''), :quirks_mode => true, :allow_nan => true)
     end
 
     def csv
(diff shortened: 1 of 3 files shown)


lib/httparty/parser.rb
```diff
@@ -114,8 +114,10 @@ def xml
       MultiXml.parse(body)
     end
 
+    UTF8_BOM = "\xEF\xBB\xBF".freeze
+
     def json
-      JSON.parse(body, :quirks_mode => true, :allow_nan => true)
+      JSON.parse(body.gsub(/^#{UTF8_BOM}/, ''), :quirks_mode => true, :allow_nan => true)
```
Rsullivan00: I'm not sure if this is the right spot for this logic, or if I should be checking encoding somehow before trying to strip the BOM.

I'm open to suggestions, if there is a better way to do this.

jnunemaker:
I really don't know much about this, but it seems like an ok fix to me. Thanks!

PR #521, opened 2017-04-24
[#516] - Remove body normalization, but do it for `setup_raw_request`

### LINKED ISSUE #516: Don't use a the query string normalizer to transform the body
Scenario:
I want to communicate with a json API. I want to to a post request with a json body and for this I want to pass a hash into the body.

```
class Abc
  include HTTParty
....
  def self.authorize
    r = post('/auth/local', body: { email: username, password: password })
  end
```
What happens is, that this hash, instead of being serialized into a json-string, it is parsed into a query string. 
https://github.com/jnunemaker/httparty/blob/master/lib/httparty/request.rb#L167

Instead, I would like to specify a Proc/Class/whatever, like for the query string, but for the body payload in order to process the value that I passed for the body into the `post` method.

Rsullivan00:
Closes #516 

Here's one way to prevent the body from being serialized as a query string. For some reason, there is a test around `setup_raw_request` that expects the query string serializer to be run on the body, so I left the call for that method.

This might be a breaking change for some users, so it would be good to up the version if this is merged. I have no idea how versions are managed, so I didn't do anything for that.

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -152,7 +152,7 @@ module HTTParty
     end
 
     def body
-      options[:body].respond_to?(:to_hash) ? normalize_query(options[:body]) : options[:body]
+      options[:body]
     end
 
     def credentials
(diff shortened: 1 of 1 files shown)


Rsullivan00:
This broke some of the CLI, so I'm going to close this for now

Issue #528, opened 2017-05-11
httparty-0.15.0.gem fail on windows


net1957:
````
ruby 2.2.6p396 (2016-11-15 revision 56800) [x64-mingw32]

Updating installed gems
Updating httparty
Fetching: httparty-0.15.0.gem (100%)
D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package.rb:388:in `symlink': symlink() function is unimplemented on this machine (NotImplementedError)
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package.rb:388:in `block (2 levels) in extract_tar_gz'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package/tar_reader.rb:65:in `each'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package.rb:365:in `block in extract_tar_gz'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package.rb:459:in `block in open_tar_gz'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package.rb:456:in `wrap'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package.rb:456:in `open_tar_gz'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package.rb:364:in `extract_tar_gz'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package.rb:345:in `block (2 levels) in extract_files'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package/tar_reader.rb:65:in `each'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package.rb:342:in `block in extract_files'
        from D:/program/dvlt/ruby/ruby22-x64/lib/ruby/site_ruby/2.2.0/rubygems/package/file_source.rb:30:in `open'
[...]

PR #529, opened 2017-05-12
remove symlink

ahorek:
https://github.com/jnunemaker/httparty/commit/6888e5343aaac2dab3aac1afa05615cdec9587f5
fixes https://github.com/jnunemaker/httparty/issues/528

Is it safe to just remove the symlink from the repository?


jnunemaker:
I think removing is safe. Honestly I'm not sure why it exists. It doesn't seem to be used and CI passes. Is 0.15.1 not installing for you? I have some who have confirmed install in prod no problem.

ahorek:
it is autogenerated by this command c_rehash generated in
https://github.com/jnunemaker/httparty/blob/master/spec/fixtures/ssl/generate.sh#L15

unfortunatelly 0.15.1 didn't solve the problem because specs with this symlink are still included in a .gem file. It affects mainly windows users, but the same issue teoreticaly can happen on linux if you're using a filesystem, that doesn't support symlinks.
We're already using this version on our production linux systems and it works just fine, thanks again for your effort!

jnunemaker:
0.15.2 is out with symlink removed.

Issue #531, opened 2017-05-12
uninitialized constant SimpleDelegator


aselder:
When running license_finder with httparty 0.15.2 we get this error. Reverting to httparty 0.13.7 fixes the issue.

```
/opt/circleci/.rvm/gems/ruby-2.2.6/gems/httparty-0.15.2/lib/httparty/response/headers.rb:3:in `<class:Response>': uninitialized constant SimpleDelegator (NameError)
	from /opt/circleci/.rvm/gems/ruby-2.2.6/gems/httparty-0.15.2/lib/httparty/response/headers.rb:2:in `<module:HTTParty>'
	from /opt/circleci/.rvm/gems/ruby-2.2.6/gems/httparty-0.15.2/lib/httparty/response/headers.rb:1:in `<top (required)>'
```

aselder:
It looks like simple delegator doesn't exist in Ruby 2.2.6 at least:
```
 ~  irb
 :001 > SimpleDelegator
NameError: uninitialized constant SimpleDelegator
	from (irb):1
	from /Users/aselder/.rvm/rubies/ruby-2.2.6/bin/irb:11:in `<main>'
```

aselder:
PR coming shortly

aselder:
Closed with #532

luckyraul:
@jnunemaker release please

PR #532, opened 2017-05-12
Add missing require to headers.rb

aselder:
The delegate library is not auto-required in Ruby 2.2.x, so we need to explicitly require it if we are using SimpleDelegator.

### DIFF
--- a/lib/httparty/response/headers.rb
+++ b/lib/httparty/response/headers.rb
@@ -1,3 +1,5 @@
+require 'delegate'
+
 module HTTParty
   class Response #:nodoc:
     class Headers < ::SimpleDelegator


aselder:
Fix for: https://github.com/jnunemaker/httparty/issues/531

aselder:
```
 ~  irb
 :001 > RUBY_VERSION
 => "2.2.6"
 :002 > SimpleDelegator
NameError: uninitialized constant SimpleDelegator
	from (irb):2
	from /Users/aselder/.rvm/rubies/ruby-2.2.6/bin/irb:11:in `<main>'
 :003 > require 'delegate'
 => true
 :004 > SimpleDelegator
 => SimpleDelegator
```

Issue #534, opened 2017-05-16
HEAD requests are broken since 0.15.0 (undefined method force_encoding for nil:NilClass)


peret:
In versions >= 0.15.0 when I do something like `HTTParty.head("https://www.google.com")`, it fails with this Callstack:
```
NoMethodError: undefined method `force_encoding' for nil:NilClass
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty/request.rb:280:in `encode_with_ruby_encoding'
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty/request.rb:316:in `_encode_body'
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty/request.rb:322:in `encode_body'
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty/request.rb:350:in `handle_response'
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty/request.rb:158:in `perform'
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty/request.rb:347:in `handle_response'
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty/request.rb:158:in `perform'
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty.rb:563:in `perform_request'
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty.rb:536:in `head'
	from /Users/peter/.rvm/gems/ruby-2.3.3@ptcshell/gems/httparty-0.15.0/lib/httparty.rb:629:in `head'
```

It works as expected in version 0.14.0. I'm fairly certain that it is caused by one of the changes in this PR: #513 
[...]

PR #535, opened 2017-05-16
Bugfix: Don't attempt to call force_encoding on nil objects.

peret:
This should fix HEAD requests, where the body will be nil.
Fixes issue #534.


jnunemaker:
Dup of https://github.com/jnunemaker/httparty/pull/530

PR #540, opened 2017-05-18
[#539] Use non-destructive gsub, fix test stubbing

### LINKED ISSUE #539: The change made in 0.15.4 breaks httparty if frozen_string_literal is set to true
The change made in 2f8093871b9bb6375f640c68145eb295733a97e2 breaks our builds because of the use of `gsub!`

```
RuntimeError:
       can't modify frozen String
     # ./vendor/bundle/ruby/2.3.0/gems/httparty-0.15.4/lib/httparty/parser.rb:105:in `gsub!'
     # ./vendor/bundle/ruby/2.3.0/gems/httparty-0.15.4/lib/httparty/parser.rb:105:in `parse'
     # ./vendor/bundle/ruby/2.3.0/gems/httparty-0.15.4/lib/httparty/parser.rb:67:in `call'
     # ./vendor/bundle/ruby/2.3.0/gems/httparty-0.15.4/lib/httparty/request.rb:383:in `parse_response'
     # ./vendor/bundle/ruby/2.3.0/gems/httparty-0.15.4/lib/httparty/request.rb:351:in `block in handle_response'
     # ./vendor/bundle/ruby/2.3.0/gems/httparty-0.15.4/lib/httparty/response.rb:25:in `parsed_response'
     # ./vendor/bundle/ruby/2.3.0/gems/httparty-0.15.4/lib/httparty/response.rb:88:in `method_missing'
```

We use Ruby 2.3.3 with the `# frozen_string_literal: true` magic comment in preparation for immutable strings in Ruby 3.0

Rsullivan00:
Closes #539

Apparently I like breaking stuff with my changes 😭 . This PR fixes `gsub!` from breaking on frozen string literals.

Some of the tests around this were actually stubbing the `body` call directly, instead of setting the data. I've changed these tests to inject the data instead. As a high-level note, a lot of tests for `Parser` are stubbing private methods, which I don't think is a good way to structure tests.

We should be able to test all use cases of `Parser` through the external API. When we stub, or test directly, internal methods, any implementation change can break any number of other tests, even if the change doesn't affect external behavior.

### DIFF
--- a/lib/httparty/parser.rb
+++ b/lib/httparty/parser.rb
@@ -102,7 +102,7 @@ module HTTParty
       return nil if body == "null"
       return nil if body.valid_encoding? && body.strip.empty?
       if body.valid_encoding? && body.encoding == Encoding::UTF_8
-        body.gsub!(/\A#{UTF8_BOM}/, '')
+        @body = body.gsub(/\A#{UTF8_BOM}/, '')
       end
       if supports_format?
         parse_supported_format
--- a/spec/httparty/parser_spec.rb
+++ b/spec/httparty/parser_spec.rb
@@ -63,57 +63,62 @@ RSpec.describe HTTParty::Parser do
   end
 
   describe "#parse" do
-    before do
-      @parser = HTTParty::Parser.new('body', :json)
-    end
-
     it "attempts to parse supported formats" do
-      allow(@parser).to receive_messages(supports_format?: true)
-      expect(@parser).to receive(:parse_supported_format)
-      @parser.parse
+      parser = HTTParty::Parser.new('body', :json)
+      allow(parser).to receive_messages(supports_format?: true)
+
+      expect(parser).to receive(:parse_supported_format)
+      parser.parse
     end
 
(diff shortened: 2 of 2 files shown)


jnunemaker:
> As a high-level note, a lot of tests for Parser are stubbing private methods, which I don't think is a good way to structure tests.

Agreed. I wasn't the maintainer when most of the parsing stuff went in (took some time away) and just haven't had the energy to clean things up. Thanks!

henrik:
I guess this is the fix for this issue?

```
07:45:09 >> HTTParty.get("https://auctionet.com/api/v2/companies.json")
Encoding::CompatibilityError: incompatible encoding regexp match (UTF-8 regexp with ASCII-8BIT string)
	from /Users/henrik/.asdf/installs/ruby/2.4.1/lib/ruby/gems/2.4.0/gems/httparty-0.15.2/lib/httparty/parser.rb:120:in `gsub'
```

Thanks for fixing it! Was just about to report it.

Issue #541, opened 2017-05-24
httparty trying to json parse string w/ invalid byte sequence


jaypatrickhoward:
Previously we were experiencing the issue in #533, so I had us pinned to 0.14.0.  Since that was fixed, I removed the pin and tested with 0.15.5.  Now I'm getting:

Exception Backtrace:  822: unexpected token at '^_�^H'

/usr/local/rvm/gems/ruby-2.2.4/gems/json-1.8.6/lib/json/common.rb:155:in `parse'
/usr/local/rvm/gems/ruby-2.2.4/gems/json-1.8.6/lib/json/common.rb:155:in `parse'
/usr/local/rvm/gems/ruby-2.2.4/gems/httparty-0.15.5/lib/httparty/parser.rb:123:in `json'
/usr/local/rvm/gems/ruby-2.2.4/gems/httparty-0.15.5/lib/httparty/parser.rb:143:in `parse_supported_format'
/usr/local/rvm/gems/ruby-2.2.4/gems/httparty-0.15.5/lib/httparty/parser.rb:108:in `parse'
/usr/local/rvm/gems/ruby-2.2.4/gems/httparty-0.15.5/lib/httparty/parser.rb:67:in `call'
/usr/local/rvm/gems/ruby-2.2.4/gems/httparty-0.15.5/lib/httparty/request.rb:383:in `parse_response'
/usr/local/rvm/gems/ruby-2.2.4/gems/httparty-0.15.5/lib/httparty/request.rb:351:in `block in handle_response'
/usr/local/rvm/gems/ruby-2.2.4/gems/httparty-0.15.5/lib/httparty/response.rb:25:in `call'
/usr/local/rvm/gems/ruby-2.2.4/gems/httparty-0.15.5/lib/httparty/response.rb:25:in `parsed_response'

When I revert to 0.14.0 this same call works.  I'll try to dig up the raw response we get, but it will take some work.

jnunemaker:
Any more info you can provide would be helpful. Without some idea of the raw response I'm not sure how to fix it.

jaypatrickhoward:
It may have been fixed with the fixes for #542 and #543.  Will test again with latest and report back.

jaypatrickhoward:
Or, since 0.15.5 is still latest, I'll wait for the next release and test with that.

jnunemaker:
0.15.6 is out. Please try and let me know.

deep-spaced:
[...]

PR #543, opened 2017-06-02
Ensure correct encoding is used for Content-Types that are lower-case or aliases

### LINKED ISSUE #542: Response bodies aren’t encoded when `charset` in header doesn’t have the right case
I’ve been having some character encoding issues with responses that *seemed* ok. For example, a response with the header:

```
Content-Type: text/plain; charset=utf-8
```

Would be encoded as `ASCII-8BIT` when I received it:

```
> response.body.encoding
=> #<Encoding:ASCII-8BIT>
```

This appears to be because [`Request#encode_with_ruby_encoding` does an exact name match](https://github.com/jnunemaker/httparty/blob/master/lib/httparty/request.rb#L279) on the list of encodings to see whether it can be converted: 

```rb
def encode_with_ruby_encoding(body, charset)
  if !body.nil? && Encoding.name_list.include?(charset)
    body.force_encoding(charset)
    ...
```

…but the name for utf-8 is `UTF-8`, not `utf-8`, as it was set in the HTTP header in my example above.

It looks like this broke in c46663d73ff96d86b92d36208a021bbe58672e7c, where the check switched from `Encoding.find(charset)` to `Encoding.name_list.include?(charset)`, I assume to avoid the rescue block. We could probably do a case-insensitive search on `name_list`, but I think that could still get false-positives or miss aliases. (I didn’t even know encoding aliases were a thing until today!) So I think the right fix here is to switch back to the rescue statement (but only rescue from `ArgumentError`).

I *think* this might be the cause behind #541 as well; the timing of the change seems right.

**I’m happy to submit a PR if desired; just let me know!**

*Side note: I thought my conclusion here was crazy at first because [there is a test for exactly this scenario](https://github.com/jnunemaker/httparty/blob/master/spec/httparty/request_spec.rb#L464-L470). It turns out the test doesn’t really work: the stubbed response is already a UTF-8 string, so when httparty fails to apply the encoding from the HTTP header, the test passes anyway because the body was already properly encoded to begin with.*

Mr0grog:
Somewhere along the line, it looks like responses stopped getting the encoding specified by the `Content-Type` header applied unless the header’s value was an exact, case-sensitive match for an encoding’s name in Ruby.

In a lot of real-world scenarios, encodings will be referenced in lower-case (internally, they are all upper-case in Ruby) or by other well-known aliases, e.g. `eucKR` is an alias for `EUC-KR` and `CP65001` is an alias for `UTF-8`.

This also updates tests—they were previously passing erroneously and didn’t *actually* test the behavior they were meant to. Test values were already correct and didn’t need the behavior the tests were looking for to be applied.

This fixes #542 and possibly #541.

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -276,11 +276,10 @@ module HTTParty
     end
 
     def encode_with_ruby_encoding(body, charset)
-      if !body.nil? && Encoding.name_list.include?(charset)
-        body.force_encoding(charset)
-      else
-        body
-      end
+      encoding = Encoding.find(charset)
+      body.force_encoding(charset)
+    rescue ArgumentError
+      body
     end
 
     def assume_utf16_is_big_endian
--- a/spec/httparty/request_spec.rb
+++ b/spec/httparty/request_spec.rb
@@ -157,46 +157,46 @@ RSpec.describe HTTParty::Request do
       expect(@request.instance_variable_get(:@raw_request)['authorization']).not_to be_nil
     end
 
-    context 'digest_auth' do 
-      before do 
+    context 'digest_auth' do
+      before do
         response_sequence = [
-          {status: ['401', 'Unauthorized' ], 
+          {status: ['401', 'Unauthorized' ],
             www_authenticate: 'Digest realm="Log Viewer", qop="auth", nonce="2CA0EC6B0E126C4800E56BA0C0003D3C", opaque="5ccc069c403ebaf9f0171e9517f40e41", stale=false',
-            set_cookie: 'custom-cookie=1234567', 
+            set_cookie: 'custom-cookie=1234567',
           },
           {status: ['200', 'OK']}
-        ]      
+        ]
         FakeWeb.register_uri(:get, "http://api.foo.com/v1",
                             response_sequence)
-      end 
+      end
 
       it 'should not send credentials more than once' do
         response_sequence = [
-          {status: ['401', 'Unauthorized' ], 
+          {status: ['401', 'Unauthorized' ],
             www_authenticate: 'Digest realm="Log Viewer", qop="auth", nonce="2CA0EC6B0E126C4800E56BA0C0003D3C", opaque="5ccc069c403ebaf9f0171e9517f40e41", stale=false',
-            set_cookie: 'custom-cookie=1234567', 
+            set_cookie: 'custom-cookie=1234567',
           },
-          {status: ['401', 'Unauthorized' ], 
+          {status: ['401', 'Unauthorized' ],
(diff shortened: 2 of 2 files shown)


lib/httparty/request.rb
```diff
@@ -276,11 +276,10 @@ def get_charset
...
 
     def encode_with_ruby_encoding(body, charset)
-      if !body.nil? && Encoding.name_list.include?(charset)
-        body.force_encoding(charset)
-      else
-        body
-      end
+      encoding = Encoding.find(charset)
+      body.force_encoding(charset)
+    rescue ArgumentError
```
jnunemaker: Could you explain the change from checking body not being nil to rescuing ArgumentError? Thanks!

Mr0grog: Ohhhhh, this is a mistake; I had assumed the test changes in c46663d73ff96d86b92d36208a021bbe58672e7c covered the `body.nil?` case that was added to this method in the same commit and then didn’t worry about this because the tests passed.

Turns out they don’t; I’ll add a test for that and put the explicit `body.nil?` check back in.

(FWIW, I forgot the `ArgumentError` type check wouldn’t cover the `body.nil?` case and was thinking leaving out the explicit `nil` check made the code a little simpler. D'oh!)

Mr0grog: Actually, it looks like this method can never get called with a `nil` body: https://github.com/jnunemaker/httparty/blob/master/lib/httparty/request.rb#L350

Do you still want an explicit `nil?` check in this method?

I can add a test anyway—it seems reasonable to have a “response body is nil/empty” test—but that test would never actually exercise this method.

Mr0grog: Oof, nevermind entirely! There’s already a test that does this and I missed it: https://github.com/jnunemaker/httparty/blob/master/spec/httparty/request_spec.rb#L488-L493

So… it seems like the `nil?` check is entirely unnecessary here *and* it’s well-covered by tests. (I’m guessing it became unnecessary sometime after c46663d73ff96d86b92d36208a021bbe58672e7c)

Mr0grog: Ok, last thing I’ll post here. It looks like there were two different solutions to the `nil` body problem added within just a few days of each other, so I’d guess one was added without knowledge of the other:

1. May 12, the check inside `encode_with_ruby_encoding` and a test: 9bd77528fddf4f20d3f5ed749d340f535ee19765
2. May 16, just never call `encode_with_ruby_encoding` at all in this case: d721b76680c2b4936a8820d065da0039f851bbcb

jnunemaker:
@Mr0grog you rock. Thanks!

Mr0grog:
No problem, happy to help! :D

Issue #510, opened 2017-02-01
Test suite does not pass on Ruby 2.4


voxik:
The test suite fails on Ruby 2.4 with errors such as:

~~~
  1) HTTParty::Request::NON_RAILS_QUERY_STRING_NORMALIZER doesn't modify strings
     Failure/Error: expect(CGI.unescape(query_string)).to eq("foo=bar&foo=baz")
     NameError:
       uninitialized class variable @@accept_charset in #<Class:CGI>
     # ./spec/httparty/request_spec.rb:13:in `unescape'
     # ./spec/httparty/request_spec.rb:13:in `block (3 levels) in <top (required)>'
~~~

Since I believe this is bug in Ruby, I reported it upstream [[1]], so this is more or less just heads up ;)

[1]: https://bugs.ruby-lang.org/issues/13178

Rsullivan00:
I have 4 different failures on ruby 2.4.1

```
Finished in 0.76511 seconds (files took 0.3414 seconds to load)
582 examples, 4 failures

Failed examples:

rspec ./spec/httparty/ssl_spec.rb:29 # HTTParty::Request SSL certificate verification should work when no trusted CA list is specified, even with a bogus hostname, when the verify option is set to true
rspec ./spec/httparty/ssl_spec.rb:33 # HTTParty::Request SSL certificate verification should work when using ssl_ca_file with a self-signed CA
rspec ./spec/httparty/ssl_spec.rb:37 # HTTParty::Request SSL certificate verification should work when using ssl_ca_file with a certificate authority
rspec ./spec/httparty/ssl_spec.rb:19 # HTTParty::Request SSL certificate verification should work when no trusted CA list is specified, when the verify option is set to false

Randomized with seed 42560
```

Issue #524, opened 2017-05-09
Request to support Ruby 2.4.0


Lucashuang0802:
It failed due to the json error: https://github.com/jnunemaker/httparty/pull/523

Lucashuang0802:
@jnunemaker

Lucashuang0802:
Possible duplicate with #510

jnunemaker:
Yep, looks like a dup of #510. Agreed that I'd like to fix whatever problems there are currently.

PR #544, opened 2017-06-03
Replaces Fakeweb with Webmock in tests.

JonMidhir:
Fakeweb hasn't had a release in years and sadly no longer appears to be maintained. As a result tests are failing on more modern Rubies (2.4.x). This change replaces Fakeweb with Webmock, which is actively maintained, making tests green on all supported Rubies.

Fixes https://github.com/jnunemaker/httparty/issues/510 and https://github.com/jnunemaker/httparty/issues/524

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ rvm:
   - 2.1.8
   - 2.2.4
   - 2.3.0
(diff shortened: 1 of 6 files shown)


jnunemaker:
Thanks! This is great.

PR #545, opened 2017-06-20
Give some spacing love to code snippets in docs/README

rbmrclo:
This might be a long shot (and a very minor one too), but while I was checking out some examples in [docs/README](https://github.com/jnunemaker/httparty/blob/master/docs/README.md), I noticed that most of them are not properly indented.

#### Some points to consider:
- The official [README.md](https://github.com/jnunemaker/httparty/blob/master/README.md) uses 2 spaces. This PR is just following the consistency for the docs.
- Might be opinionated, but 2-spaces is already considered as a [community standard](https://github.com/bbatsov/ruby-style-guide#spaces-indentation).
- This will reduce the cognitive load when reading the code
- When copying the code snippets to our favorite editors (probably for those who haven't discovered auto indent), it requires another action for the developer to trim the spaces just to make it consistent with the two spaces code.

#### Before 
httpartybad

#### After
httpartygood

Please review. Thanks!

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -24,11 +24,10 @@ You can use this guide to work with SSL certificates.
 # Use this example if you are using a pem file
 
 class Client
-	include HTTParty
-	
-	base_uri "https://example.com"
-	pem File.read("#{File.expand_path('.')}/path/to/certs/cert.pem"), "123456"
+  include HTTParty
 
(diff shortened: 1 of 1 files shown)


jnunemaker:
Thanks! No idea how the spacing got off on those.

PR #546, opened 2017-06-24
capture cookies returned from http_digest setup request

heythisisnate:
This change captures the cookies returned from the original http_digest 401 request, so that they can be used on subsequent requests. The use case here is that the session cookie is only returned on the first request.

### DIFF
--- a/spec/httparty/request_spec.rb
+++ b/spec/httparty/request_spec.rb
@@ -226,6 +226,14 @@ RSpec.describe HTTParty::Request do
         expect(raw_request.get_fields('cookie')).to eql ["custom-cookie=1234567"]
       end
 
+      it 'should capture cookies returned from a 401 response' do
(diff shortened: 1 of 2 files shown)


spec/httparty/request_spec.rb
```diff
@@ -226,6 +226,14 @@
         expect(raw_request.get_fields('cookie')).to eql ["custom-cookie=1234567"]
       end
 
+      it 'should capture cookies returned from a 401 response' do
```
jnunemaker: Could you explain what is different about this test than the one prior `it 'should maintain cookies returned from a 401 response'`?

jnunemaker:
Closing this as stale since I didn't get a response from @heythisisnate, but if you see this at any point and could respond to my question above I'd be happy to help move this forward.

PR #548, opened 2017-07-14
Add ResponseError response attribute to message

MrLeebo:
When I raise a HTTParty::ResponseError, the log message is not very helpful.

# Repro

```rb
raise HTTParty::ResponseError.new(response)
```

# Expected

```
HTTParty::ResponseError: { "status"=>"error", "message"=>"That's not how this works, that's not how any of this works!" }
```

# Actual

```
HTTParty::ResponseError: HTTParty::ResponseError
```

# Resolution

So I added the response to the error message.


jnunemaker:
Thanks for caring!

Issue #381, opened 2015-03-10
response.class and response.instance_of?


CBluowei:
```ruby
response = Httpaty.get(url)

response.class #==> HTTParty::Response,

#but...

response.kind_of?(HTTParty::Response) #==> false
response.kind_of?(Hash) #==> true
```

why do I get this case??

wolfascu:
This is due to the way [`HTTParty::Response` implements `method_missing`](https://github.com/jnunemaker/httparty/blob/2d2779e7555f2a0927d79ba7fdb37921c3310dd9/lib/httparty/response.rb#L65).

The call to `method_missing` is delegated to the internal `parsed_response` and `response` objects. The former is a `String`, and it's the one that responds to `is_a?`/`kind_of?`.

Therefore, 

```ruby
response.kind_of? String #===> true
```

jnunemaker:
@wolfascu is correct. Its an awkward part of httparty due to keeping backwards compatibility when we added lazy parsing.

JonMidhir:
Is there anything to be said for implementing `HTTParty::Response#is_a?`? I ran into this as well and it caused a bit of a problem. I'm happy to issue the PR of course.

cameronwp:
I ran into this bug yesterday and it was super confusing. IMO, backwards compat should not outweigh meeting reasonable expectations for the behavior `instance_of?` / `kind_of?` / `is_a?`. I'd definitely +1 an effort to fix this.

JonMidhir:
@cameronwp I actually addressed this in https://github.com/jnunemaker/httparty/pull/444 but it's a bit of a hack tbh.

Ultimately anything to get around the `method_missing` override is going to be, though.

I'd started work a while back to refactor this whole class and remove the override. I can perform some necromancy on that if the maintainers think it's worthwhile. Would end backwards compatibility but we could put an update message in the gemspec to warn.

/cc @jnunemaker

cameronwp:
My mistake! Looks like we've been using v0.13.7.

jnunemaker:
@JonMidhir any thoughts on an option for this that people can opt into? Maybe also a deprecation warning and then eventually we flip flop the option? That seems like an ok path forward. HTTParty is used a lot, so I'm not sure I'm comfortable with only a message in the changelog or what not.

JonMidhir:
[...]

PR #550, opened 2017-08-02
Deprecates method delegation via method_missing

JonMidhir:
Delegating methods to other objects using method_missing in instances
of HTTParty::Response is causing unexpected behaviour and has resulted
in multiple issues. Work-arounds to specific issues have been
successful but this change starts the process of removing the root
cause.

Discussion at https://github.com/jnunemaker/httparty/issues/381

- The user will receive a deprecation warning when using a method that is delegated.
- To remove the warning they should call the method on `response` or `parsed_response` instead.
- To opt-in to the new behaviour globally the user can call `HTTParty::Response.disable_method_delegation!`

Most of this code is temporary, to facilitate the transition from one set of
behaviours to another.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,11 @@
+## 0.15.7
+
+Change
+
+* [Deprecated method delegation to response](https://github.com/jnunemaker/httparty/pull/550)
+
+## 0.15.6
+
 ## 0.15.5
 
 Fixed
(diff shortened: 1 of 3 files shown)


jnunemaker:
Closing as stale. I think I'd be open to a version of this with no conflicts.

PR #558, opened 2017-10-03
Silence some noisy spec warnings in Ruby 2.4

diversiforge:
Ruby 2.4 caused the specs to become slightly noisier due to a deprecation warning for `Fixnum` and a warning in Ruby's `reject` (see https://github.com/ruby/ruby/blob/v2_4_1/hash.c#L1335-L1337 ). This PR makes a couple of small tweaks to eliminate the new warnings.


jnunemaker:
Thanks! You rock!

PR #564, opened 2017-12-16
Add socks proxy support

jdrago999:
I looked through some old issues and discovered #87 which even had a PR.

HTTParty has changed internally since then, so I took parts of that PR and made it work against the current master.

I'm doing this for a personal project and won't be adding tests.

There is an example added as part of this PR.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -13,22 +13,22 @@
     * Creates a custom parser for XML using crack gem
     * Uses `get` request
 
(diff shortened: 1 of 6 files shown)


jnunemaker:
Did a quick fly by and it looks good. I don't know much about socks so I'm not comfortable merging this without tests, but if someone else wants to take a stab at reviving this, I'd be up for merging.

PR #569, opened 2018-01-30
Add multipart support

TheSmartnik:
Hi,
This pull request fixes long-standing issue, of not handling file upload properly #52. No extra dependencies introduced. Tested file upload on rails server, works alright.


jnunemaker:
Neat. I really don't know much about multipart uploads, but this seems simple and contained so I'm going to merge it. @TheSmartnik are you ok with me pinging you in the future if questions or other pull requests come up about this?

TheSmartnik:
@jnunemaker sure, I'd be super happy to help

jirutka:
This implementation is broken – header `Content-Type` is not added when you pass _any_ header into the request method.

TheSmartnik:
@jirutka oh, I see. Thanks a lot for feedback and pr

wayne530:
May be worth emphasizing for many folks that decided to use `httmultiparty` when on `httparty` < 0.16 that using >= 0.16 + `httmultiparty` causes issues and multi-part form data isn't properly encoded.

PR #572, opened 2018-02-20
Add support for Object#tap to HTTParty::Response

matrinox:
This is more complicated than just writing one myself but I feel this will keep the behaviour consistent across Ruby versions


matrinox:
nvm, you guys have fixed this in a later version that we're not on

PR #575, opened 2018-02-24
Fix multipart uploads with boolean params.

rywall:
I am using a gem ([ocr_space](https://github.com/suyesh/ocr_space)) that [passes a boolean param](https://github.com/suyesh/ocr_space/blob/master/lib/ocr_space/convert.rb#L17) to `HTTParty.post`. This started failing in v0.16 with the following error message:

```ruby
TypeError: no implicit conversion of false into String
from /gems/httparty-0.16.0/lib/httparty/request/body.rb:40:in `+'
```


jnunemaker:
I'm going to merge this. I have some hesitation adding the to_s as it could cover up passing in nil or other not serializable objects accidentally, but we'll see how it goes. :)

PR #579, opened 2018-02-28
Fix rspec warnings

JonMidhir:
This change cleans up some of the warnings output when running test suite.

- Removes the redefining of `MyParser` class in individual tests.
- Instantiates all variables that are used while tests are running with sensible defaults (they're actually fine to be `nil`, so long as they're defined).
- Removes a specific error check that could create false positives.

An error with the SSLTestServer support object related to unknown CAs isn't fixed by this.


jnunemaker:
@JonMidhir thanks! Really appreciate PRs like this.

Issue #574, opened 2018-02-22
Hashes nested inside of arrays not properly parsed in query params / body.


sealabcore:
Example hash payload that does not parse properly:

```ruby
{:marketing_event => { :marketed_resources=>[{:type=>"product", :id=>57474842640}]}}
```

The resulting string is `marketing_event[marketed_resources][][type]=product`. You can see that the `id` is completely omitted from the resulting string.

This could have been a result of the refactor of this class in the latest version.
https://github.com/jnunemaker/httparty/blob/master/lib/httparty/hash_conversions.rb

jnunemaker:
Thanks for the report. I'd be up for a PR. I can try to tackle it at some point, but I'm not sure when. If anyone else has time, let me know.

jmortlock:
Hi @jnunemaker I have add a PR which fixes this, let me know if you need anything extra to get this merged.

jnunemaker:
Closed by #583

PR #583, opened 2018-03-19
Correctly handle nested params

jmortlock:
Had the same problem as was mentioned in Issue #574. 

The normalize_param expected an array structure of
``` [ [key_a, value_a], [key_b, value_b] ]```

However when recursing you would get structure like this, 
```[ [key_a, value_a, key_b, value_b] ]```


ddefrenne:
Hi, @jnunemaker could you release a new version? We're having the same problem at the moment. Thanks in advance :).

Issue #584, opened 2018-03-22
upload file give an error


metripurari:
I am still getting error whle uploading the file

The parameter looks like this
             {"file"=>
           #<ActionDispatch::Http::UploadedFile:0x007ffedaf3e068
          @content_type="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
         @headers=
         "Content-Disposition: form-data; name=\"attachment[file]\"; filename=\"gstr1-FEB 2018.xlsx\"        \r\nContent-Type: application/vnd.openxmlformats-officedocument.spreadsheetml.sheet\r\n",
         @original_filename="gstr1-FEB 2018.xlsx",
       @tempfile=#<File:/var/folders/c_/t9781sl51gz1yfcf2fqfz6p40000gn/T/RackMultipart20180322-71273-1xq1kos>>,
      "reference_id"=>19}

When I try to upload it through 

        HTTParty.post("#{base_url}?token=#{token}",
                                      body: @params,
                                      detect_mime_type: true
                                      )

It give an error 

      TypeError (no implicit conversion of ActionDispatch::Http::UploadedFile into String)

jnunemaker:
What version of rails? What version of httparty? Could you get a backtrace of the exception? My gut says it isn't detecting `ActionDispatch::Http::UploadedFile` as an uploaded file even though it should. If so, maybe [this line](https://github.com/jnunemaker/httparty/blob/db8df8c58bbb7182fe52c20eb64924b5f3e70ac1/lib/httparty/request/body.rb#L39) is where the conversion is failing?

mcls:
Hi, I just ran into the same issue.

It seems related to the following lines:

https://github.com/jnunemaker/httparty/blob/db8df8c58bbb7182fe52c20eb64924b5f3e70ac1/lib/httparty/request/body.rb#L32-L35

`file?(value)` returns true for `ActionDispatch::Http::UploadedFile`, but when you call `File.basename` on it, it fails with:

```
*** TypeError Exception: no implicit conversion of ActionDispatch::Http::UploadedFile into String
```

mcls:
If it's changed too `File.basename(path)` it would work with both File/Tempfile and ActionDispatch::Http::UploadedFile:

```
(byebug) File.basename(value.path)
"RackMultipart20180327-53658-2lj8b3.jpg"
(byebug) File.basename(Tempfile.new.path)
"20180327-53658-13pgemv"
```

[...]

PR #585, opened 2018-03-27
Support ActionDispatch::Http::UploadedFile again

mcls:
Using the `path` method instead of passing the file directly to `File.basename` ensures that `ActionDispatch::Http::UploadedFile` can be used again.

Related issue: https://github.com/jnunemaker/httparty/issues/584

Note: ActionDispatch isn't a dependency of the library, so I opted not to write a test specifically for it. I tried to account for it in the existing tests by checking that e.g. `to_str` isn't called on the file object, but that didn't work since `File.basename` is implemented in C.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* [Support ActionDispatch::Http::UploadedFile again](https://github.com/jnunemaker/httparty/pull/585)
+
 ## 0.16.1
 
 * [Parse content with application/hal+json content type as JSON](https://github.com/jnunemaker/httparty/pull/573)
--- a/lib/httparty/request/body.rb
+++ b/lib/httparty/request/body.rb
@@ -32,7 +32,7 @@ module HTTParty
         multipart = normalized_params.inject('') do |memo, (key, value)|
           memo += "--#{boundary}\r\n"
           memo += %(Content-Disposition: form-data; name="#{key}")
-          memo += %(; filename="#{File.basename(value)}") if file?(value)
+          memo += %(; filename="#{File.basename(value.path)}") if file?(value)
           memo += "\r\n"
           memo += "Content-Type: application/octet-stream\r\n" if file?(value)
           memo += "\r\n"


jnunemaker:
So this works with regular files and action dispatch uploads? Is path right? File.basename returns the filename, not a full path, but path makes me think it is more than just the filename. Any thoughts either way on that? Just curious.

mcls:
Hi, it's still using [`File.basename`](https://ruby-doc.org/core-2.5.0/File.html#method-c-basename). What I meant to explain is that passing the `ActionDispatch::Http::UploadedFile` object directly to `File.basename` doesn't work, while it does for regular `File` objects.

All I changed is passing the file's `path` to the object instead, which is what `File.basename` does normally. The effect is the same, it's just that it supports ActionDispatch now.

Example:

```ruby
irb(main):003:0> f = File.open("lib/httparty.rb")
=> #<File:lib/httparty.rb>
irb(main):004:0> f.path
=> "lib/httparty.rb"
irb(main):005:0> File.basename(f)
=> "httparty.rb"
irb(main):006:0> File.basename(f.path)
=> "httparty.rb"
```

versus

```ruby
irb(main):002:0> file = ActionDispatch::Http::UploadedFile.new(tempfile: File.open("config/routes.rb"))
#<ActionDispatch::Http::UploadedFile:0x00007fe148b67bd8 @tempfile=#<File:config/routes.rb>, @original_filename=nil, @content_type=nil, @headers=nil>
irb(main):003:0> file.path
"config/routes.rb"
irb(main):004:0> File.basename(file.path)
"routes.rb"
irb(main):005:0> File.basename(file)
TypeError: no implicit conversion of ActionDispatch::Http::UploadedFile into String
        from (irb):5:in `basename'
        from (irb):5
        from /Users/mcls/.rbenv/versions/2.3.5/lib/ruby/gems/2.3.0/gems/railties-5.0.6/lib/rails/commands/console.rb:65:in `start'
        from /Users/mcls/.rbenv/versions/2.3.5/lib/ruby/gems/2.3.0/gems/railties-5.0.6/lib/rails/commands/console_helper.rb:9:in `start'
        from /Users/mcls/.rbenv/versions/2.3.5/lib/ruby/gems/2.3.0/gems/railties-5.0.6/lib/rails/commands/commands_tasks.rb:78:in `console'
        from /Users/mcls/.rbenv/versions/2.3.5/lib/ruby/gems/2.3.0/gems/railties-5.0.6/lib/rails/commands/commands_tasks.rb:49:in `run_command!'
        from /Users/mcls/.rbenv/versions/2.3.5/lib/ruby/gems/2.3.0/gems/railties-5.0.6/lib/rails/commands.rb:18:in `<top (required)>'
        from bin/rails:4:in `require'
        from bin/rails:4:in `<main>'
```

jnunemaker:
> Hi, it's still using File.basename.

I overlooked that somehow. 😆 My bad.

jnunemaker:
This is out in 0.16.2. Thank you so much for caring and fixing this.

Yedfree (approved):

PR #588, opened 2018-04-24
stream_body fragment include http_response feature

shanempope:
Add a `:stream_body_include_response_code` option to get a response code back with the fragment when streaming data. This allows for skipping redirects while streaming (and other error detection), which before would tack the redirect onto the file.

We needed this, I appreciate any suggestions.

### DIFF
--- a/examples/stream_download.rb
+++ b/examples/stream_download.rb
@@ -9,8 +9,14 @@ url = "https://cdn.kernel.org/pub/linux/kernel/v4.x/#{filename}"
 
 File.open(filename, "w") do |file|
   response = HTTParty.get(url, stream_body: true) do |fragment|
-    print "."
-    file.write(fragment)
+    if [301, 302].include?(fragment.code)
+      print "skip writing for redirect"
+    elsif fragment.code == 200
+      print "."
+      file.write(fragment)
+    else
+      raise StandardError, "Non-success status code while streaming #{fragment.code}"
+    end
   end
 end
 puts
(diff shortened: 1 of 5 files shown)


shanempope:
How we use this:
```
  url = 'SOMETHING THAT REDIRECTS using 302'
  HTTParty.get(url, stream_body: true, stream_body_include_response_code: true) do |resp_fragment|
    file.write(fragment) if resp_fragment[:code] == 200
  end```

shanempope:
I seriously appreciate you taking the time to look at my pull request. I quite like your approach. I am too inexperienced in ruby to understand the ramifications of making it a forwardable, and any breaking effects it could have on your users. I assume, if it's possible to do something a certain way in ruby, someone, somewhere is using your library that way.

I made the changes and pushed them. Let me know what you think.
-shane

shanempope:
@jnunemaker Do you think this will ever make it in? Thanks

TheSmartnik:
@shanempope sorry for the delay. I'll look into it later today

TheSmartnik:
@shanempope everything seems fine, thanks for the pr. Can you please add example usage to [examples directory](https://github.com/jnunemaker/httparty/tree/master/examples) so this feature wouldn't be lost

shanempope:
@TheSmartnik Thanks, I added an example and fixed `FragmentWithResponse.code` to be an int to match the way HTTParty's response.code is.

TheSmartnik:
@shanempope thanks! Looks good

PR #596, opened 2018-06-29
Added support for lock and unlock http requests

Oldharlem:


jnunemaker:
Can't say I've ever used this, but why not. 😄 Thanks!

PR #598, opened 2018-07-02
Fix multipart filename for ActionDispatch::Http::UploadedFile TempFile

zherr:
Closes #585


zherr:
I'll CC @TheSmartnik here as they seem active in reviewing PRs as of late. I would greatly appreciate your input!

TheSmartnik:
Hi @zherr ,
I don't how qualified I'm to review your pr here, but PR looks great, thanks!
RestClient solves the problem the same way for 9 years *(should've looked at it when implemented the feature 😩)*, already. So I think we are good.

The only thing that bothers me is `expose_dsl_globally`. It looks very much unnecessary here.

zherr:
Thanks for taking a look @TheSmartnik.

As for `expose_dsl_globally`, this is purely a testing configuration for rspec, and won’t have any affect outside of running tests. Docs can be seen here: https://relishapp.com/rspec/rspec-core/v/3-8/docs/configuration/global-namespace-dsl

I’m hoping this one could get merged and released. @jnunemaker , when you have a few minutes, would you reviewing  this one?

zherr:
@TheSmartnik check out the latest commit: https://github.com/jnunemaker/httparty/pull/598/commits/30d90368f65d376abe64cabe58feff3a1afb42e1

I've made the use of the RSpec DSL consistent, to avoid the possible confusion as you outlined.

PR #601, opened 2018-08-09
gemspec homepage does not exist

notEthan:
this PR points the gemspec homepage at a page that exists - or if the page just got lost somewhere, restoring it would resolve this issue.


jnunemaker:
@notEthan you rock! Thanks.

PR #606, opened 2018-08-17
Process dynamic headers before making actual request

ajitsing:
I had a requirement where I had to pass `Request-ID` header in all the http calls. So I tried using `HTTParty.headers()` method to achieve this. But while testing I realized that once the value is set in headers using `HTTParty.headers()`, it will remain as is.

But I want different behaviour where the value is evaluated every time the call is fired.

One idea is to accept block as header value which can be evaluated when the request is fired. I have solved it in my project by adding a patch to httparty. I thought it would be useful to others as well.

Example of dynamic header:

```ruby
require 'rails_distributed_tracing'

class BaseApi
  include HTTParty
  headers DistributedTracing::REQUEST_HEADER_KEY => lambda {DistributedTracing.current_request_id}
end
```

Awaiting for suggestion and feedback.

### DIFF
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -571,12 +571,10 @@ module HTTParty
     end
 
     def process_dynamic_headers(headers)
-      headers = headers.map do |header, value|
-        value = value.respond_to?(:call) ? value.call : value
-        [header, value]
+      headers.each_with_object({}) do |header, processed_headers|
+        key, value = header
+        processed_headers[key] = value.respond_to?(:call) ? value.call : value
       end
-
-      Hash[headers]
     end
 
     def process_cookies(options) #:nodoc:


lib/httparty.rb
```diff
@@ -566,9 +566,19 @@ def perform_request(http_method, path, options, &block) #:nodoc:
     def process_headers(options)
       if options[:headers] && headers.any?
         options[:headers] = headers.merge(options[:headers])
+        options[:headers] = process_dynamic_headers(options[:headers])
       end
     end
 
+    def process_dynamic_headers(headers)
+      headers = headers.map do |header, value|
```
TheSmartnik: What you want here is [Enumerable#each_with_object](https://ruby-doc.org/core-2.5.1/Enumerable.html#method-i-each_with_object). Resulting code will much cleaner, than map + conversion to hash

ajitsing: @TheSmartnik thanks for the suggestion, changed the code accordingly.

Issue #595, opened 2018-06-29
HTTParty multipart doesn't support changing of content-type


pangtongs:
In using multipart of HTTParty, the content-type of the file we opened will be hardcoded to be  using `Content-Type: application/octet-stream`.

However, there are cases where we want to send data with the different content-type, let's say we want to specify a jpg image and thus requires it to be `Content-Type: image/jpg`.

This case is currently not supported by the HTTParty system.

jnunemaker:
Can you provide an example of how you are attempting to use content type?

zherr:
👋 Just stumbled across this issue when making the jump from [httmultiparty](https://github.com/jwagener-soundcloud/httmultiparty) (now no longer maintained) to HTTParty's support for multipart requests.

We ran into an issue where an API we communicate with is expecting the `Content-Type` of the `multipart/form-data` data in the request to be set to the known media type. For example, uploading a `.png` file, the client would set the `Content-Type` to `image/png`.

We wanted to make sure this wasn't just a bug in the API we were using, so we did a little research. Specifically, we were using [this API call](https://developer.atlassian.com/cloud/jira/platform/rest/#api-api-2-issue-issueIdOrKey-attachments-post), which references [RFC 1867](https://www.ietf.org/rfc/rfc1867.txt), which was suspiciously experimental. We found a more updated specification ([RFC 2388](https://www.ietf.org/rfc/rfc2388.txt)) that confirms the expected behavior of the client:

>If the contents of a file are returned via filling out a form, then the file input is identified as the appropriate media type, if known, or "application/octet-stream".

So after all that, HTTParty should set the appropriate `Content-Type` of a multipart request if it knows how to. 

Overall this looks like a great addition to HTTParty. I'm guessing this new behavior should be implemented in a way that is backward-compatible and adds an option to determine the mime type. 

What does everyone else think?

[...]

PR #608, opened 2018-08-24
Fix multipart to set its header even when other headers are provided without overwriting provided content type

meuble:
In regards to issues opened about [Content Type issue with multipart requests](https://github.com/jnunemaker/httparty/issues/595), also mentioned [here](https://github.com/jnunemaker/httparty/pull/569) in comments by @jirutka, there was a [fix implemented](https://github.com/jnunemaker/httparty/pull/576).

I think this fix is incorrect as it prevent any `Content Type` customisation: if the request is multipart, it is always overwritten.
I propose a fix here that allow for `Content Type` customisation, but also makes sure that if we have custom headers but no `Content Type`, the content type is set to `multipart/form-data`.

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -205,6 +205,16 @@ module HTTParty
       @raw_request = http_method.new(request_uri(uri))
       @raw_request.body_stream = options[:body_stream] if options[:body_stream]
 
+      if options[:body]
+        body = Body.new(options[:body], query_string_normalizer: query_string_normalizer)
+        if body.multipart?
+          content_type = "multipart/form-data; boundary=#{body.boundary}"
+          options[:headers] ||= {}
+          options[:headers]['Content-Type'] ||= content_type
+        end
+        @raw_request.body = body.call
+      end
+
       if options[:headers].respond_to?(:to_hash)
         headers_hash = options[:headers].to_hash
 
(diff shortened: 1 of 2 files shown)


meuble:
We may want to include @TheSmartnik in the discussion to assess the viability of this pull request.

TheSmartnik:
@meuble I've looked at the issue you've mentioned. Thanks for the pr, but It seems the problem in the issue you've mentioned in this line https://github.com/jnunemaker/httparty/blob/master/lib/httparty/request/body.rb#L39. I simply hardcoded `octet-stream`, without detecting content-type of the file

meuble:
@TheSmartnik That might be another issue.
The fact that we can't customise the Content Type header for requests with attachment remains, unless with the fix in my pr.

meuble:
Yes.
Please refer to the following documentation http://rasa.com/docs/nlu/master/http/
As they specify:

> The request should always be sent as application/x-yml regardless of wether you use json or md for the data format. Do not send json as application/json for example.

To do so, the HTTParty code I should write to query this API is the following

```
response = HTTParty.post("#{BASE_URL}/train", 
  body: {:file => File.open("intent_utterances.md")}, # Yes, it's a markdown file, here. Not a Yaml...
  headers: {'Content-Type' => "application/x-yml"}
)
```

In the current state of HTTParty, the Content Type that is sent to the API is `multipart/form-data` and I get an error. After my PR, the Content Type is `application/x-yml` and every thing is fine.
I agree that HTTParty should extract the content type from the file, if none is provided.
But even if HTTParty does it, I still want to have the freedom to specify a custom Content Type.

TheSmartnik:
@meuble I see. Thanks!

You don't need multipart in this case. Following documentation, what you need to do is to send file directly. With httparty, you would do so, by streaming content of the file. Like so

```ruby
HTTParty.put(
  'http://localhost:5000/train',
  body_stream: File.open('sample_configs/config_train_server_md.yml', 'r'),
  headers: { 'Content-Type' => 'application/x-yml', 'Transfer-Encoding' => 'chunked' }
)
```

This should produce expected payload.

---

Uploading files and uploading files through multipart is two different things. If we encode body params as multipart, but change content-type to *anything* else. Server won't be able to decode it

meuble:
Ok, I wasn't aware of `body_stream` as a way to add files to HTTP requests.
I close the pull request. Sorry for the inconvenience...

PR #609, opened 2018-10-07
Unify loggers

springerigor:
This PR introduces following changes:

1. Use `public_send` instead of `send` for calling logger's public method.
2. Unify naming of a method returing formatted `Time.now` (`current_time` vs `time`).
3. Adjust `ApacheFormatter` to be more similar to `CurlFormatter` (private methods instead of multiple local variables, log message extracted to a dedicated method).


jnunemaker:
Thanks!

PR #610, opened 2018-10-08
Adds editorconfig

harman28:
Noticed some unattended whitespace lying around the codebase. Inferred an editorconfig and applied it wherever I could with minor changes. 

Also added the config file to make things a little easier for future potential contributors.


jnunemaker:
Thanks!

Issue #472, opened 2016-05-02
Content-Type Does not send when I try to use a new Hash syntax on header


kidush:
### It's work fine

``` ruby
{ 'Content-Type' => 'application/json' }
```
### but  when I try to use this way:

``` ruby
{ 'Content-Type': 'application/json' }
```

the header doesn't send :(

jnunemaker:
> the header doesn't work :(

Does it error? Just not send? Can you provide more details please. There isn't much I can do to help without a small script that shows what you expect and what happened.

kidush:
Sorry, @jnunemaker I've changed the title 😄 

for example: 

```ruby
class Request
  include HTTParty
  # this way does not send the header
  headers 'Content-Type': 'application/json'
  # this way works
  headers 'Content-Type' => 'application/json'
end
```

Other thing that I've seen, in ruby 2.3 the request works fine with `{ 'Content-Type': 'application/json' }` in ruby 2.2 does not work.

In Ruby 2.2 there is a method `capitalize` in Net::HttpHeader specifically at line 172 
that is different from Ruby 2.3.

`/net/http/header.rb`
```ruby
# ruby 2.2
def capitalize(name)
  name.split(/-/).map {|s| s.capitalize }.join('-')
end

# ruby 2.3
def capitalize(name)
  name.to_s.split(/-/).map {|s| s.capitalize }.join('-')
end
```

jnunemaker:
I've actually never used `headers 'Content-Type': 'application/json'` syntax. I didn't even know it worked. I thought you always need to do a string with the =>, a symbol with => or keyword with :. Does it work if you do `headers :'Content-Type' => 'application/json'`? Note the ":" before content type forcing it to be a symbol. Probably will. I'm wondering if this is just a ruby hash syntax thing or a net/http thing and if there is anything httparty can or should do. What are your thoughts?

kule:
@jnunemaker this caught me out actually - where I was setting the ````Content-Type: 'application/json'```` but the post request ended up being changed to ````application/x-www-form-urlencoded```` instead.  I put it down to being too used to HashWithIndifferentAccess! :)

The new "string": symbol syntax appeared in ruby 2.2 I believe.

Since no one had replied I thought I'd try out your suggestion - for me it didn't work with ````headers :'Content-Type' => 'application/json'```` either, which is to be expected I guess, if it only works with strings and not symbols.

[...]

PR #622, opened 2018-11-10
Stringify headers

TheSmartnik:
FIxes https://github.com/jnunemaker/httparty/issues/472

### DIFF
--- a/spec/httparty_spec.rb
+++ b/spec/httparty_spec.rb
@@ -214,18 +214,16 @@ RSpec.describe HTTParty do
       end
     end
 
-    if RUBY_VERSION >= '2.2.0'
-      context 'when headers passed as symbols' do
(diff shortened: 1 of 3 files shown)


spec/httparty_spec.rb
```diff
@@ -213,6 +213,19 @@ def expect_headers(header = {})
         @klass.post('http://example.com', body: { file: File.open('spec/fixtures/tiny.gif')})
       end
     end
+
+    context 'when headers passed as symbols' do
+      let(:headers) { { 'Content-Type' => 'application/json', 'Custom-Header' => 'example' } }
+
+      it 'convers them to string' do
```
jnunemaker: `convers` => `converts`

TheSmartnik:
> Any downside to this?

@jnunemaker I don't think there is any. All of the other clients I've checked convert symbol headers to string without hesitation

PR #626, opened 2018-11-30
HTTParty::Request::Body#multipart? fixes

EmilioCristalli:
This PR intends to fix a few issues with the `multipart?` method:

1. It wasn't working for params that respond to `#to_hash` but not to `#detect` (like `ActionDispatch::Request::Session`). Same for params that respond to `#to_ary` but not to `#detect`
```ruby
class HashLike; def to_hash; { a: 3 }; end; end
HTTParty::Request::Body.new(a: HashLike.new).multipart? # NoMethodError: undefined method `detect' for #<HashLike:0x00007fbf8971ae08>
```

2. Wasn't detecting files inside arrays
(`includes_hash?(value)` was always returning `true` for not empty arrays, because all objects respond to `#hash`, and then it was doing `hash.detect do |key, value| ` with an array which actually yields only one arg)
```ruby
HTTParty::Request::Body.new(a: [Tempfile.new(['some_temp_file','.gif'])]).multipart? # false
```

3. looked like it was intended to be recursive and look for a file inside all hash values or array elements recursively but I think it wasn't doing it (if the method was not supposed to be recursive please let me know because this might impact performance)

### DIFF
--- a/lib/httparty/request/body.rb
+++ b/lib/httparty/request/body.rb
@@ -22,7 +22,7 @@ module HTTParty
       end
 
       def multipart?
-        params.respond_to?(:to_hash) && (force_multipart || has_file?(params.to_hash))
+        params.respond_to?(:to_hash) && (force_multipart || file?(params))
       end
 
       private
@@ -46,24 +46,14 @@ module HTTParty
         multipart += "--#{boundary}--\r\n"
       end
 
-      def has_file?(hash)
-        hash.detect do |key, value|
-          if value.respond_to?(:to_hash) || includes_hash?(value)
-            has_file?(value)
-          elsif value.respond_to?(:to_ary)
-            value.any? { |e| file?(e) }
-          else
-            file?(value)
-          end
+      def file?(value)
+        if value.respond_to?(:to_hash)
+          value.to_hash.any? { |_, v| file?(v) }
+        elsif value.respond_to?(:to_ary)
+          value.to_ary.any? { |v| file?(v) }
+        else
+          value.respond_to?(:path) && value.respond_to?(:read)
         end
-      end
-
-      def file?(object)
-        object.respond_to?(:path) && object.respond_to?(:read) # add memoization
-      end
-
-      def includes_hash?(object)
-        object.respond_to?(:to_ary) && object.any? { |e| e.respond_to?(:hash) }
       end
 
       def normalize_query(query)
--- a/lib/httparty/request/body.rb
+++ b/lib/httparty/request/body.rb
@@ -22,7 +22,7 @@ module HTTParty
       end
 
       def multipart?
-        params.respond_to?(:to_hash) && (force_multipart || file?(params))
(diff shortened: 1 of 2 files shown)


FLarra:
It'd be great to have this PR merged as soon as you feel confident with it. 

I'm maintaining the gem https://github.com/smartinez87/exception_notification which relies on HTTParty in several notifiers and they are not working if updating HTTParty version. The last working version is on tag: https://github.com/jnunemaker/httparty/releases/tag/v0.15.7

It'd be great to merge this #multipart? fix so we don't need to lock our users to versions prior to the multipart refactor.

Cheers

lib/httparty/request/body.rb
```diff
@@ -46,26 +46,16 @@ def generate_multipart
...
-      def has_file?(hash)
-        hash.detect do |key, value|
-          if value.respond_to?(:to_hash) || includes_hash?(value)
-            has_file?(value)
-          elsif value.respond_to?(:to_ary)
-            value.any? { |e| file?(e) }
-          else
-            file?(value)
-          end
+      def file?(value)
```
TheSmartnik: I feel like `#has_file?` is better a naming for this method, anyway.
You still ask whether the `params` have a file or not. You're right that it can be `file`, but it can also be hash, array and `#has_file?` is a more inclusive naming

lib/httparty/request/body.rb
```diff
@@ -46,26 +46,16 @@ def generate_multipart
...
+        if value.respond_to?(:to_hash)
+          value.to_hash.any? { |_, v| file?(v) }
+        elsif value.respond_to?(:to_ary)
+          value.to_ary.any? { |v| file?(v) }
+        else
+          value.respond_to?(:path) && value.respond_to?(:read)
```
TheSmartnik: Previous use of a method was more readable and understandable , imho.
It showed that if value responds to `path` and `read` it is a file, which is exactly what we are looking for.

lib/httparty/request/body.rb
EmilioCristalli: yeah, you are right. I changed here 5a778ed2eac655ecebc0e2b134e9d53a29026aa9

lib/httparty/request/body.rb
EmilioCristalli: you mean this method?

```ruby
      def file?(object)
        object.respond_to?(:path) && object.respond_to?(:read)
      end
```

if that's what you meant, i think it makes sense, and also would help for the `generate_multipart` which should only check if the value is a file (not recursive) if i understand it correctly.

I changed it here as well: 5a778ed2eac655ecebc0e2b134e9d53a29026aa9

EmilioCristalli:
@TheSmartnik  thanks for the review!
I think i addressed your comments, but let me know if i misunderstood them or if you have any other comments!

TheSmartnik:
Thanks! Look good. I'll run a few more tests later today and bump the version

PR #635, opened 2018-12-11
Do not stream the body of redirects

will3216:
Currently, if you use the following code, the body of the redirect is written to the provided block in the request.perform method, and there is no external way to differentiate the redirect body from the destination body.
```ruby
res = []
HTTParty.get('http://example.com') do |data|
  res << data
end
```
For instance, with the above example, imagine the get request to http://example.com returns a page with a status of 302, and a body of "Redirecting to some other page". "Redirecting to some other page" is going to get written to the block that you provided, even though it's probably not what you want and there is currently no way to tell that the data came from the redirect.

Thus, this adds the `ignore_redirect_body` option which allows you to skip the body of the redirects.

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -386,11 +386,12 @@ module HTTParty
     end
 
     def response_redirects?(response = nil)
-      case response || last_response
+      _response = response || last_response
+      case _response
       when Net::HTTPNotModified # 304
         false
       when Net::HTTPRedirection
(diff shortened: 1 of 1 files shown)


shanempope:
https://github.com/jnunemaker/httparty/pull/588

will3216:
@shanempope Looks like your branch already has this covered. Awesome! I'll go ahead and close this PR as not needed

Issue #555, opened 2017-09-15
Handle BOM characters in text/csv response


andreyprostakov:
I send POST request with JSON body and receive response with content type "text/csv" and BOM characters (\xEF\xBB\xBF) at the beginning. I cannot access some of its attributes, including content_type:
```
result = HTTParty.post(url, parameters)
result.body # "\xEF\xBB\xBF\"Column A\",\"Column B\"\r\n\"A1\",\"B1\""
result.content_type # CSV::MalformedCSVError: Illegal quoting in line 1
CSV.parse(result.body) # CSV::MalformedCSVError: Illegal quoting in line 1
```

A bug?

As a workaround, I enforce `format: :plain`, but that's not pretty in my case.
I use httparty 0.15.6, ruby 2.2.3.

jnunemaker:
This was recently an issue with json as well. https://github.com/jnunemaker/httparty/issues/498 I know nothing about it really, but I'm open to a PR.

TheSmartnik:
@andreyprostakov thank you for the issue and sorry for a delay in response.

I managed to reproduce  it but the problem appeases to be in a server that sends you a file. HTTParty always tries to remove BOM characters in UTF-8 responses, so the only reason it doesn't do it is because your response is not UTF-8 _(It's probably unspecified at all)_

Unless you can change backend of the server, the only possible workaround right now is the one you are using.

PR #641, opened 2019-01-24
Always try to remove BOM characters

TheSmartnik:
Following https://github.com/jnunemaker/httparty/issues/555#issuecomment-457013866, I thought that it may be useful to try to always remove BOM chars, because not every server properly specifies charset.


jnunemaker:
@TheSmartnik I still don't really get the BOM stuff so do what you think is right. If things blow up we can always revert and release a new version.

jnunemaker:
Closing as stale.

PR #644, opened 2019-02-03
[WIP] Fix encoding of streamed chunk

### LINKED ISSUE #643: Encoding Incorrect when Streaming Response
I'm running into an issue where the encoding is not set according to the response HTTP headers when using `:stream_body => true`. 

I've provided a minimal reproducible example in a gist at https://git.io/fhSBB.

In short, if I look at `req.headers['content-type']` I (correctly) see `["text/html;charset=UTF-8"]` and `req.body.encoding` returns `UTF-8`. However, for whatever reason, `fragment.encoding` (`fragment` being the argument to the block provided to `HTTParty.get(URL, :stream_body => true)` ) is `ASCII-8BIT`.

Oddly, it works fine when not using streaming.

I'm happy to help troubleshoot this if there's any way I can be of help.  Any assistance would be appreciated.

TheSmartnik:
Fix #643

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,6 @@ gemspec
 gem 'rake'
 gem 'mongrel',  '1.2.0.pre2'
 
-
 group :development do
   gem 'guard'
   gem 'guard-rspec'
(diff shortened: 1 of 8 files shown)


jnunemaker:
@TheSmartnik Looks like a pry require failure. Probably just need the gemfile tweaked on ci to include pry. I think right now pry is only in dev.

TheSmartnik:
@jnunemaker Thanks!

PR #647, opened 2019-02-18
Add support for `Net::HTTP#write_timeout` method (Ruby 2.6.0)

springerigor:
[The method](https://ruby-doc.org/stdlib-2.6/libdoc/net/http/rdoc/Net/HTTP.html#method-i-write_timeout-3D) was introduced in Ruby 2.6.0.

The gem already supports `open_timeout` & `read_timeout` options, so it would be nice to provide support for `write_timeout` as well.

From the official documentation:

> Number of seconds to wait for one block to be written (via one write(2) call). Any number may be used, including Floats for fractional seconds. If the HTTP object cannot write data in this many seconds, it raises a Net::WriteTimeout exception. The default value is 60 seconds. Net::WriteTimeout is not raised on Windows.

### DIFF
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -173,9 +173,9 @@ module HTTParty
     #     include HTTParty
     #     default_timeout 10
     #   end
-    def default_timeout(t)
-      raise ArgumentError, 'Timeout must be an integer or float' unless t && (t.is_a?(Integer) || t.is_a?(Float))
-      default_options[:timeout] = t
+    def default_timeout(value)
+      validate_timeout_argument(__method__, value)
+      default_options[:timeout] = value
     end
 
     # Allows setting a default open_timeout for all HTTP calls in seconds
@@ -184,9 +184,9 @@ module HTTParty
     #     include HTTParty
     #     open_timeout 10
     #   end
-    def open_timeout(t)
-      raise ArgumentError, 'open_timeout must be an integer or float' unless t && (t.is_a?(Integer) || t.is_a?(Float))
-      default_options[:open_timeout] = t
+    def open_timeout(value)
+      validate_timeout_argument(__method__, value)
+      default_options[:open_timeout] = value
     end
 
     # Allows setting a default read_timeout for all HTTP calls in seconds
@@ -195,9 +195,9 @@ module HTTParty
     #     include HTTParty
     #     read_timeout 10
     #   end
-    def read_timeout(t)
-      raise ArgumentError, 'read_timeout must be an integer or float' unless t && (t.is_a?(Integer) || t.is_a?(Float))
-      default_options[:read_timeout] = t
+    def read_timeout(value)
+      validate_timeout_argument(__method__, value)
+      default_options[:read_timeout] = value
     end
 
     # Allows setting a default write_timeout for all HTTP calls in seconds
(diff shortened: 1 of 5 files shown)


jnunemaker:
Seems good to me. @TheSmartnik any thoughts either way?

lib/httparty.rb
```diff
@@ -200,6 +200,19 @@ def read_timeout(t)
...
 
+    # Allows setting a default write_timeout for all HTTP calls in seconds
+    # Supported by Ruby > 2.6.0
+    #
+    #   class Foo
+    #     include HTTParty
+    #     write_timeout 10
+    #   end
+    def write_timeout(t)
+      raise ArgumentError, 'write_timeout must be an integer or float' unless t && (t.is_a?(Integer) || t.is_a?(Float))
```
TheSmartnik: Could extract it to method and refactor other timeout methods as well, please?
Something like `validate_timeout_argument(timeout)` or `check_timeout_argument(timeout)`

lib/httparty/connection_adapter.rb
```diff
@@ -126,6 +133,14 @@ def connection
         http.open_timeout = options[:open_timeout]
       end
 
+      if options[:write_timeout] && (options[:write_timeout].is_a?(Integer) || options[:write_timeout].is_a?(Float))
```
TheSmartnik: There is a lot of duplication here. Would be better to extract it to a private method `add_timeout?(options[:write_timeout])`

lib/httparty/connection_adapter.rb
```diff
@@ -116,6 +117,12 @@ def connection
       if options[:timeout] && (options[:timeout].is_a?(Integer) || options[:timeout].is_a?(Float))
         http.open_timeout = options[:timeout]
         http.read_timeout = options[:timeout]
+
+        if RUBY_VERSION >= "2.6.0"
+          http.write_timeout = options[:timeout]
+        else
+          Kernel.warn("Warning: option :write_timeout requires Ruby version 2.6.0 or later")
```
TheSmartnik: It's a timeout for everything, we should set `write_timeout` if we can, but there is no need to add warning here.

lib/httparty/connection_adapter.rb
```diff
@@ -126,6 +133,14 @@ def connection
         http.open_timeout = options[:open_timeout]
       end
 
+      if options[:write_timeout] && (options[:write_timeout].is_a?(Integer) || options[:write_timeout].is_a?(Float))
+        if RUBY_VERSION >= "2.6.0"
```
TheSmartnik: There is a lot of duplication as well. Probably a would be better, to create a wrapper method like so _(just an idea, not sure about the naming)_
```ruby
for_versions_above('2.6', option: :write_timeout) do
  http.write_timeout = options[:write_timeout]
end
```

spec/httparty/connection_adapter_spec.rb
```diff
@@ -191,6 +200,19 @@
           allow(Net::HTTP).to receive_messages(new: http)
           adapter.connection
         end
+
+        it "should not set the write_timeout" do
+          http = double(
+            "http",
+            :null_object => true,
```
TheSmartnik: Please, use colon notation for symbol keyword arguments

springerigor: Hey @TheSmartnik thanks for the comments 👍 I addressed all of them, but this one. I would like to discuss one thing here. 

Due to the fact that the doubles contain names with special characters (e.g. `:use_ssl= => false`) converting them to Ruby 1.9 hash syntax would require using `'` to wrap them. Not sure if it is worth doing in this case. I am totally fine with leaving the current syntax.

Besides that, feel free to take a look at the pushed commits. The code looks better now IMO.

springerigor:
@TheSmartnik A kind reminder for taking another look 👋

TheSmartnik:
@springerigor thanks a lot! Sorry for a long reply

PR #648, opened 2019-02-19
Expose .connection on fragment block param

### LINKED ISSUE #633: Any way to access the certificate used for the connection?

machty:
This provides access to the connection object,
which in turn exposes properties like `.peer_cert`
for performing additional validation against
x509 certificates.

Closes #633, Follow-up to #634

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -81,3 +81,6 @@
 
 * [Uploading File](body_stream.rb)
   * Uses `body_stream` to upload file
+
+* [Accessing x509 Peer Certificate](peer_cert.rb)
+  * Provides access to the server's TLS certificate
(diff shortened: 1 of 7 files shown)


machty:
@TheSmartnik Ready for your review. Wondering if `FragmentWithResponse` needs a new name given that it now also has a `connection` object?

machty:
@TheSmartnik Hi, just checking in for an update.

machty:
@TheSmartnik Pinging for feedback :)

TheSmartnik:
@machty thanks! Great job!

I'll rename the class myself before releasing a new version, so no worries

PR #656, opened 2019-04-12
Add blank? method to check for when response & body is nil or empty.

NickDiMucci:
This PR is to add a `blank?` method to the `Response` class. Currently, `Response` overrides `nil?` as so

https://github.com/jnunemaker/httparty/blob/e1421556f6e0a9d86c1026a14a3969c7a382a92b/lib/httparty/response.rb#L77

This is fine, however, I feel that if you're going to override `nil?`, you should also override `blank?` to behave similarly. Currently, `.blank?` will return false when `nil?` is true due to `response.body.empty?` returning true, which is confusing and can lead to nasty, hard to track bugs.

For example, if you call `HTTParty.get("https://httpbin.org/status/204")`, this will return a response with an empty body and `response.blank?` returns false. I believe it should return true.

### DIFF
--- a/lib/httparty/response.rb
+++ b/lib/httparty/response.rb
@@ -78,6 +78,10 @@ module HTTParty
       response.nil? || response.body.nil? || response.body.empty?
     end
 
+    def blank?
+      nil?
+    end
+
     def to_s
       if !response.nil? && !response.body.nil? && response.body.respond_to?(:to_s)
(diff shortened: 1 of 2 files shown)


NickDiMucci:
Since `blank?` is technically a Rails method, I better understand why it's not a part of the library, and thus I'm going to abandon the PR.

Issue #655, opened 2019-04-11
Exception being raised when reading http_version on unparseable responses


slicedpan:
When reading the `http_version` of a response, this is delegated to `parsed_response` through a `method_missing` call which raises an exception if the response is not parseable.

pjmartorell:
What is the issue? Raising an exception is not a bug itself, is your responsibility to rescue and recover from an exception if the response is not the expected

slicedpan:
Sorry, just realised I should have been a bit clearer. In my use case, the HTTP response is valid, but the body is not parseable according to the reported content-type. 

Calling `http_version` on such a response throws an exception, even though the HTTP part of the response is valid.

pjmartorell:
If you could post some code and the full trace of the exception, maybe we could see where is the problem

slicedpan:
Hi, I just submitted a pull request. The issue was due to checking the `parsed_response` first in `method_missing`, before checking the `Net::HTTP` response. 

I'm not 100% sure about the spec tbh, it's a bit messy but it seems like it would be hard to write a spec to cover this without pulling apart a lot of the current response spec. Let me know if you have any suggestions.

PR #657, opened 2019-04-17
Unparseable body exception fix

slicedpan:
This fixes issue #655

### DIFF
--- a/spec/httparty/response_spec.rb
+++ b/spec/httparty/response_spec.rb
@@ -39,9 +39,15 @@ RSpec.describe HTTParty::Response do
     it "should set code" do
       expect(@response.code).to eq(@response_object.code)
     end
-
+        
     it "should set code as an Integer" do
       expect(@response.code).to be_a(Integer)
+    end
+    
+    it "should set http_version" do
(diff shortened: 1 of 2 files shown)


spec/httparty/response_spec.rb
```diff
@@ -39,10 +39,16 @@
...
-
+        
     it "should set code as an Integer" do
       expect(@response.code).to be_a(Integer)
     end
+    
+    it "should set http_version" do
+      unparseable_body = lambda { raise "Unparseable" }
+      unparseable_response = HTTParty::Response.new(@request_object, @response_object, unparseable_body)
+      expect(unparseable_response.http_version).to eq(@response_object.http_version)
```
jnunemaker: It seems that the following:

```ruby
expect(unparseable_response.response.http_version).to eq(@response_object.http_version)
```

Also works fine. The problem is that I can understand how delegation would be expected to work as well. As you said, it doesn't because parsed_response is checked first in method missing. I'm tempted to change the order, but no idea how severe of a change that would be, so I'll just merge this.

slicedpan: :+1:

Issue #421, opened 2015-07-14
Parser error raised immediately prevents access to response details


GeReV:
I'm currently working with a slightly *problematic* HTTP API, which on occasion returns non-JSON responses on errors.
I.e. a 500 error with a simple "not available" plaintext response.

Such a case raises a `JSON::ParserError` and hides away details that could be relevant for error handling, such as the status code returned:

```ruby
begin
  response = HTTParty.get('url.json', ...) # Raises a JSON::ParserError
rescue
  # response is unavailable to check response.code
end
```

I think a solution for this would be to perform the parsing lazily, when first trying to access `parsed_response` or any other data that is actually derived from the body.

An example for such a use case:
```ruby
response = HTTParty.get('url.json', ...) # Does not raise an exception

puts response.code # response.code is available

begin
  response.parsed_response # Raises the parser exception here
resuce
  ...
end
```

jnunemaker:
What version are you on? Pretty sure that parsing is lazy.

https://github.com/jnunemaker/httparty/blob/78da14d4f6ccae82f7727f2d682e117713fdb9d0/lib/httparty/request.rb#L289
https://github.com/jnunemaker/httparty/blob/78da14d4f6ccae82f7727f2d682e117713fdb9d0/lib/httparty/response.rb#L22-L24

GeReV:
I am currently on version 0.13.3. Was this added at a later point?

jnunemaker:
It should be in 0.13. I see the psuedo code, but any chance you could provide a small script that show the error to be sure it is httparty and not something in your app? Thanks!

GeReV:
I managed to get the following to throw:
```ruby
HTTParty.get('http://getstatuscode.com/503', headers: { 'Content-Type' => 'application/json' }, format: :json)
```

Looks like I forgot to add the `format: :json` bit. :\

jnunemaker:
I put this into testing.rb:

```ruby
require "rubygems"
require "httparty"

response = HTTParty.get('http://getstatuscode.com/503', headers: { 'Content-Type' => 'application/json' }, format: :json)
p response.code

begin
  response.parsed_response
rescue => exception
  p 'bombed'
  p exception.class
end
```

I then ran it:

```
$ ruby testing.rb 
503
"bombed"
JSON::ParserError
```

That was on 0.13.5, but I would think any 0.13 would be fine. Note that if you do anything in irb it will bomb as irb calls inspect and Response#inspect calls parsed body.

[...]

PR #659, opened 2019-07-08
Handle invalid JSON in Response#inspect

henrik:
See #421.

There are a few things I'm unsure about here, but see it at a starting point for a discussion:

* Can this happen for other formats than JSON as well, and do we want to handle those? (Personally I'd be happy with this one for now, and we can support more as the need arises.)

* Since this is not a very integrated test, I'm not super confident about whether this will work as intended IRL. Is there a better place or way I could test it? Or does this seem a reasonable way to test it?

### DIFF
--- a/lib/httparty/response.rb
+++ b/lib/httparty/response.rb
@@ -50,7 +50,15 @@ module HTTParty
 
     def inspect
       inspect_id = ::Kernel::format "%x", (object_id * 2)
-      %(#<#{self.class}:0x#{inspect_id} parsed_response=#{parsed_response.inspect}, @response=#{response.inspect}, @headers=#{headers.inspect}>)
+
+      parsed_response_or_body =
+        begin
+          "parsed_response=#{parsed_response.inspect}"
+        rescue JSON::ParserError
+          "parsed_response=<invalid>, body=#{body.inspect}"
+        end
+
+      %(#<#{self.class}:0x#{inspect_id} #{parsed_response_or_body}, @response=#{response.inspect}, @headers=#{headers.inspect}>)
     end
 
     CODES_TO_OBJ = ::Net::HTTPResponse::CODE_CLASS_TO_OBJ.merge ::Net::HTTPResponse::CODE_TO_OBJ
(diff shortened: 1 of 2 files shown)


TheSmartnik:
Thanks for the pr! I'll take a closer look at in the weeked and get back to you

lib/httparty/response.rb
```diff
@@ -50,7 +50,15 @@ def tap
 
     def inspect
       inspect_id = ::Kernel::format "%x", (object_id * 2)
-      %(#<#{self.class}:0x#{inspect_id} parsed_response=#{parsed_response.inspect}, @response=#{response.inspect}, @headers=#{headers.inspect}>)
+
+      parsed_response_or_body =
+        begin
+          "parsed_response=#{parsed_response.inspect}"
+        rescue JSON::ParserError
```
jnunemaker: My gut reaction is we don't want to rescue this specific error here. We allow custom parsers and they could really raise anything. That made me think maybe we just do a full rescue printing out "parsing the response raised an error" or something, but then you lose insight into the error unless you include the error info in that text. 

I also thought maybe we just removed parsed_response here but I suspect a good portion of people use inspect to see the parsed response which rules that idea out.

Since we have a Parser class, we could allow the implementation of that to handle errors. This feels like more work than I have time for right now or you might be interested in and might be breaking. Really the Parser should return some sort of response object instead of just the parsed data. That would allow us to denote success or failure and then inspect could use success/failure to determine if parsed response could be shown. 

Maybe I'm overthinking it. Open to any other ideas.

henrik: Thanks so much for the comments! That makes good sense.

Maybe a reasonable middle ground would be to make it non-JSON-specific by having a blanket rescue like you suggest. That would fix the real-world issue my team's run into and we won't break any use cases that rely on the parsed response being part of the inspect.

I don't know this codebase well currently, but I would guess that going the "parser response object" route risks becoming a bit of a rabbit hole, and could possibly impact any third party format handlers, if that's a thing. It does sound a bit more elegant, but I worry about it becoming a "perfect as the enemy of the good" situation.

henrik: Just a ping since I was reminded of this PR :)

jnunemaker:
Closing as stale.

> Maybe a reasonable middle ground would be to make it non-JSON-specific by having a blanket rescue like you suggest.

I'm 👍 to this if you or anyone wants to tackle it. Sorry this dropped off my radar for so long.

PR #661, opened 2019-07-09
Pass options to dynamic block headers

wojw5:
I encountered the following problem: API required checksum based on the request body.

This simple change allows using request options in block header like in this example:

```ruby
headers 'X-SIGNATURE' => lambda { |options|
          Digest::SHA256.hexdigest(options[:body] + configuration.api_secret)
        }
```

### DIFF
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -17,6 +17,7 @@ require 'httparty/logger/logger'
 require 'httparty/request/body'
 require 'httparty/response_fragment'
 require 'httparty/text_encoder'
+require 'httparty/headers_processor'
 
 # @see HTTParty::ClassMethods
 module HTTParty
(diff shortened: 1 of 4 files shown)


lib/httparty.rb
```diff
@@ -599,14 +599,18 @@ def process_headers(options)
...
         end
 
-        options[:headers] = Utils.stringify_keys(process_dynamic_headers(options[:headers]))
+        options[:headers] = Utils.stringify_keys(process_dynamic_headers(options))
       end
     end
 
-    def process_dynamic_headers(headers)
-      headers.each_with_object({}) do |header, processed_headers|
+    def process_dynamic_headers(options)
```
TheSmartnik: Maybe it would be better to move it to a separate class? Something like `HeaderProcessor`. I don't really like that we do everything in a base class. Additionally, the method became more complex, so it's justified

wojw5: @TheSmartnik please preview code with extracted HeadersProcessor class

lib/httparty/headers_processor.rb
```diff
@@ -0,0 +1,30 @@
...
+      return unless options[:headers]
+
+      options[:headers] = headers.merge(options[:headers]) if headers.any?
+      options[:headers] = Utils.stringify_keys(process_dynamic_headers)
+    end
+
+    private
+
+    def process_dynamic_headers
+      options[:headers].each_with_object({}) do |header, processed_headers|
```
TheSmartnik: Just, fyi. You can do `|(key, value), processed_headers|`

TheSmartnik:
@wojw5 thanks! I like what you did here, looks nice

wojw5:
@TheSmartnik Do you think this is enough to bump version and do gem release? I have another gem that depends on this feature and i will release it to RubyGems once this dependency will be on RubyGems as well.

PR #660, opened 2019-07-08
Add max_retries support

dilumn:
`Net::HTTP` had this weird behaviour of retry once on timeout for `GET` & `PUT` requests. On Ruby 2.5.0 `Net::HTTP` introduced retry configuration option https://github.com/ruby/ruby/pull/1654

Found out that `HTTParty` doesn't support it yet. I added `max_retries` option pass to `Net::HTTP` from `ConnectionAdapter`

PR #674, opened 2019-09-24
Point to correct PR URL for `max_retries`

jplhomer:
Just a small update to point to https://github.com/jnunemaker/httparty/pull/660 for `max_retries` information - looks like an incorrect one got copy/pasted in the changelog.


TheSmartnik:
Thanks!

Issue #568, opened 2018-01-25
Please don't implement `.nil?`


patrick-minton:
Ran into this today:

```
[2] pry(some_code)> response
=> nil
...
[8] pry(some_code)> response == nil
=> false
[9] pry(some_code)> response.nil?
=> true
```

Had to do this because we were assuming that code like this would cover the `nil` case:

```
raise SomeError unless response
```

Took me a bit of digging to find out that HTTParty::Response implements `nil?`:

https://github.com/jnunemaker/httparty/blob/master/lib/httparty/response.rb#L58

I don't think this is a good idea. It is widely expected behavior in ruby that `nil?` isn't overwritten, since `nil` is a singleton, so these are almost universally treated as equivalent:

```
some_object == nil
some_object.nil?
```

This is important because if they're equivalent, you can use "truthiness" to check if the object *itself* is `nil`, as opposed to if the response's *body* is `nil`, which is a different thing entirely. The definition of `nil?` above even includes an empty string in the response body, which is, again, not at all the same as when the response itself is `nil`.

This method really should be `.blank?` or some such.

hydrogen18:
This was carried over by mself when I refactored the `Response` class. I didn't see it as a great idea but was trying not to just yank the rug out from people using this library. I've sat through similar conversations with co-workers as your pry sessions where the code is baffling `nil?` but  also an instance of something other than `NilClass`

Your points are valid, but we should bump a significant version number( 15->16 here I think) since this is going to break lots of folks code.

jnunemaker:
I think I'm in favor of this. It very likely could break a lot of people's use but feels like the right thing. I believe I added nil? (if it was me) way back in the day when things were a bit more wild west.

TheSmartnik:
@jnunemaker In favor as well, but I think we should add a deprecation warning first.
I'll create a pr for this later today

jaredlt:
A note for future troubleshooters: In my case it wasn't a call to `response.nil?` directly but rather I was using Nokogiri:

**Old code**

```ruby
response = HTTParty.get('https://www.google.com')
page = Nokogiri::HTML(response) # this used to work
```

**Fix**

```ruby
page = Nokogiri::HTML(response.body) # update to this
```

viktorradnai:
Another note for future troubleshooters: logging the response object will trigger the depreciation warning as well:
```
require "httparty"
require 'logger'

logger = Logger.new($stdout)

response = HTTParty.get('https://www.google.com')
[...]

PR #676, opened 2019-10-24
Consider parsed_response when checking nil?

brendo:
This PR expands `Response.nil?` to _also_ consider the `parsed_response` before determining the result. I noticed there were no tests for the `nil?` behaviour, so I added them to help prevent regressions.

I tend to agree with https://github.com/jnunemaker/httparty/issues/568, but this PR fixes the current code to behave more predictably with JSON API's.

---

This all came about because our project updated from HTTParty 0.14 to 0.17.1 and broke our application in production. 

Our code called a JSON API which returned "null" when it no results were found. We used `response.nil?` to determine when to do a fallback, otherwise we carried on as if we had a JSON object.

In doing a deep dive it turned out that introducing the `nil?` method in https://github.com/jnunemaker/httparty/pull/513/commits/e8d1ecf6268730dc1e95ec7d03e3ddb4359dc911 (v0.15.0) changed this behaviour. Previously I believe `nil?` would be delegated to `method_missing?` which would try `parsed_response` first, then fallback to `response`. 

The `Parser` already handles the case of `'null'` so `response.parsed_response` will be set to `nil`:

https://github.com/jnunemaker/httparty/blob/f2b431454fe50c64ddd42fb411338a7aa92df94b/lib/httparty/parser.rb#L102-L104

However, because the new implementation of `response.nil?` doesn't check `parsed_response`, the result of calling `nil?` is false, because the _original_ body string is `null` which is neither nil nor blank 😅

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* [`Response.nil?` now considers the parsed response]()
+
 ## 0.17.1
 
 * [Pass options to dynamic block headers](https://github.com/jnunemaker/httparty/pull/661)
--- a/lib/httparty/response.rb
+++ b/lib/httparty/response.rb
@@ -79,7 +79,8 @@ module HTTParty
     end
 
     def nil?
-      response.nil? || response.body.nil? || response.body.empty?
+      response.nil? || response.body.nil? || response.body.empty? ||
+      parsed_response.nil? || parsed_response.empty?
     end
 
     def to_s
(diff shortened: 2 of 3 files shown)


jnunemaker:
@TheSmartnik I would be inclined to do #568 as that feels the most explicit and least confusing long term. What are your thoughts? We could definitely do a version bump and drop it in the changelog as a breaking change. I'd also be tempted to do that instead of merging this (even though I really appreciate the debugging here and opening of a PR).

TheSmartnik:
@brendo the PR is nice thanks. I really appreciate your effort in doing it. However, I agree with @jnunemaker that we better implement #568 instead

brendo:
:wave: I agree, #568 is the better solution so I'm happy to close this. I think both would be a breaking change so it's probably more sustainable to break it once with the longer term fix 👍

PR #678, opened 2019-11-28
Support gzip/deflate transfer encoding when explicit headers are set.

znep:
httparty relies on Net::HTTP's built in transparent support for gzip and deflate
transfer encoding, however that did not work if you specified your own explicit headers
because calling Net::HTTPHeader#initialize_http_header overwrites the work done in
Net::HTTPGenericRequest#initialize to set the default User-Agent, Accept,
and Accept-Encoding.  This also removes the need to duplicate the logic in
Net::HTTP around not decoding the response if the caller has explicitly set
the Accept-Encoding for their own purposes.

This does introduce a slight incompatible change in behavior where
previously specifying any custom headers would omit the default headers,
while with the new behavior you can override the values however
can't omit them entirely.

While I'm here, also remove the test for HEAD requests added in
4797c7696dfcb05c9a5b2ad4748bba90c3883263 because it was not properly
testing what it claimed to and the code it was trying to test was
removed in 6f6bf6b726484eaf50e190769bbe14c9841a2c64 anyway.

### DIFF
--- a/features/handles_compressed_responses.feature
+++ b/features/handles_compressed_responses.feature
@@ -2,7 +2,7 @@ Feature: Handles Compressed Responses
 
   In order to save bandwidth
   As a developer
-  I want to uncompress compressed responses
+  I want to leverage Net::Http's built in transparent support for gzip and deflate content encoding
 
   Scenario: Supports deflate encoding
     Given a remote deflate service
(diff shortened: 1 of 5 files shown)


jnunemaker:
Overlooked this somehow. Thanks for the research and the fix. This looks ok to me. @TheSmartnik any opinions on this prior to me merging?

PR #685, opened 2020-02-06
Support edge case cookie format with a blank attribute

jeisen:
Akamai, whether due to a bug or intentional change, started to send HTTP responses with Set-Cookie headers that appear to violate the Cookie RFC by including a blank parameter. Traditional browsers appear to handle the change without issue, but HTTParty throws an exception when copying this cookie from the response to a subsequent request.

Sample invalid header:
```
Set-Cookie: somekey=someval; domain=mydomain.com; path=/; ; Secure
```

This change silently drops the invalid parameter and treats the remainder of the header as valid.


jnunemaker:
Part of me doesn't want to include fixes for incorrect things, but this feels pretty minor so I'm :cool: with it. Thanks for contributing!

PR #687, opened 2020-02-20
cookie_hash.rb cleanup

henrik:
Just a small thing I saw when looking at the gem diff.

Renamed "value" to "data" so we don't have two things named "value" in the same context.

Possibly "k" and "v" would be better since it's more consistent with `to_cookie_string`.


TheSmartnik:
Looks good! Thanks!

Yedfree (approved):

Issue #582, opened 2018-03-16
HTTParty cannot follow redirects with relative paths containing ".."


vaneyckt:
Hey all,

given a webserver that has an endpoint of the form `<my_url>/foo/bar/` that redirects to `<my_url>/foo/bar` using a relative redirect containing "..", we notice that this redirect cannot be followed.

For example, imagine that <my_url>/foo/bar/ responds with

```
< HTTP/1.1 301
< X-Frame-Options: SAMEORIGIN
< Location: ../bar
< Set-Cookie: wd-alt-sessionid=; Path=/; Secure; HttpOnly
< Content-Length: 0
< Date: Thu, 15 Mar 2018 09:55:03 GMT
< ...
```

then HTTParty will throw a 404 when trying to redirect.

```
require 'httparty'

HTTParty.get('http://localhost:4567/foo/bar/').response.code
=> 404
```

Note that this redirect is successfully handled by both `curl` and Python's `request` library. Section 5.2.4 2A of the [RFC](https://tools.ietf.org/html/rfc3986#section-5.2.4) also states that this is valid.

jmahoney:
I've tried replicating this with a simple sinatra app and HTTParty seems to work fine.

code:
```
require 'sinatra'

get '/landing_service.html' do
  return "Service Response"
end

get '/landing/service' do
  headers['Location'] = '../landing_service.html'
  status 301
end
```

verbose curl output
```
curl -v http://localhost:4567/landing/service
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 4567 (#0)
> GET /landing/service HTTP/1.1
> Host: localhost:4567
> User-Agent: curl/7.58.0
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< Content-Type: text/html;charset=utf-8
< Location: ../landing_service.html
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
< Content-Length: 0
<
* Connection #0 to host localhost left intact
```

httparty output
```
irb(main):002:0> HTTParty.get('http://localhost:4567/landing/service').response.code
=> "200"
```

That said, there's [an incomplete cucumber test](https://github.com/jnunemaker/httparty/blob/master/features/supports_redirection.feature#L15) that I'd like to flesh out if it's helpful?

jmahoney:
I've learned a bit more by working on the cucumber test

[...]

PR #690, opened 2020-03-03
Follow absolute path when redirect location contains ".." (#582)

jmahoney:
This is a fix for #582. First time contributor so keen for feedback.

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -88,15 +88,19 @@ module HTTParty
 
     def uri
       if redirect && path.relative? && path.path[0] != "/"
-        last_uri_host = @last_uri.path.gsub(/[^\/]+$/, "")
+        if path.path.start_with?('./')
+          last_uri_host = @last_uri.path
+        else
+          last_uri_host = @last_uri.path.gsub(/[^\/]+$/, "")
+        end
 
         path.path = "/#{path.path}" if last_uri_host[-1] != "/"
         path.path = last_uri_host + path.path
 
-        # use absolute path [#582]
-        if path.path.include?('/../')
+        if path.path.include?('/../') || path.path.include?('/./')
           path.path = File.expand_path(path.path)
         end
+
       end
 
       if path.relative? && path.host
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -92,6 +92,11 @@ module HTTParty
 
         path.path = "/#{path.path}" if last_uri_host[-1] != "/"
         path.path = last_uri_host + path.path
+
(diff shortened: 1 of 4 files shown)


lib/httparty/request.rb
```diff
@@ -92,6 +92,11 @@ def uri
 
         path.path = "/#{path.path}" if last_uri_host[-1] != "/"
         path.path = last_uri_host + path.path
+
+        # use absolute path [#582]
+        if path.path.include?('/../')
```
TheSmartnik: Should we account for paths like `./foo/bar` as well?

lib/httparty/request.rb
```diff
@@ -92,6 +92,11 @@ def uri
 
         path.path = "/#{path.path}" if last_uri_host[-1] != "/"
         path.path = last_uri_host + path.path
+
+        # use absolute path [#582]
```
TheSmartnik: I think this comment is unnecessary, people could you `git blame` to figure out why this change was added

spec/httparty/request_spec.rb
```diff
@@ -664,6 +664,18 @@
           expect(response.parsed_response).to eq({"hash" => {"foo" => "bar"}})
         end
 
+        it "redirects to an absolute path if a 300 contains a relative path in the location header" do
+          redirect = stub_response '', 300
+          redirect['location'] = '../bar/baz'
+          ok = stub_response('<hash><bar>baz</bar></hash>', 200)
+          allow(@http).to receive(:request).and_return(redirect, ok)
+          response = @request.perform
```
TheSmartnik: This needs to be surrounded by newlines from above and below to visually separate the test into 3 sections _Arrange, Act, Assert_

lib/httparty/request.rb
jmahoney: It's a slightly different case but happy to tackle it if you think it's worth it?  

Given a request to https://test.com/v1
 
a redirect to a relative path of "../foo/bar" should result in a new request to https://test.com/foo/bar - the foo directory is a sibling of /v1

a redirect to a  relative path of "./foo/bar" should result in a new request to https://test.com/v1/foo/bar - the foo directory is a child of v1

spec/httparty/request_spec.rb
jmahoney: Done

lib/httparty/request.rb
jmahoney: Removed

jnunemaker:
@jmahoney I dropped the ball on this. Apologies. Happy to merge if you can get it up to date. Closing as stale in the meantime.

PR #709, opened 2020-09-25
Multipart/Form-Data: rewind files after read

### LINKED ISSUE #708: File Uploads: request should rewind files after read
When making a request with files (i.e. multipart/form-data), the read head of the file object is at the end of the file.

It would be nice to rewind the file in case the client needs further use out of the object.

gi:
Fixes #708


TheSmartnik:
Nice, good catch. Thanks!

PR #710, opened 2020-10-03
Adding 2.7 and updating previous ruby versions in travis.yml

ritikesh:
Some of the rubies in the list are pretty old and EOL. Should those be removed?


TheSmartnik:
@ritikesh thanks!

Issue #143, opened 2012-05-30
Can't Marshal.dump HTTParty response for caching


ericboehs:
Since version 0.8.3 we can no longer serialize HTTParty responses for caching. I have narrowed it down to a single [commit](https://github.com/jnunemaker/httparty/commit/9fd5259c8dab00e426082b66af44ede2c9068f45). :)

I believe this is because of the lambda { } on line 211 of request.rb (Procs/lambdas can't be serialized). Of course removing the lambda breaks the call on line 18. If this can be worked around I believe we have a fix.

I have also wrote a test if it helps any:

```
it "should be serailizable (for memcaching)" do                                                                                                                                                         
  FakeWeb.register_uri(:get, 'http://twitter.com/statuses/public_timeline.xml', :body => 'tweets')                                                                                                      
  custom_parsed_response = @klass.get('http://twitter.com/statuses/public_timeline.xml')                                                                                                                
  lambda do                                                                                                                                                                                             
    Marshal.dump(custom_parsed_response)                                                                                                                                                                
  end.should_not raise_error                                                                                                                                                                            
end
```

I'm not sure how to fix this so I leave it up to you.

Thanks

rusikf:
I agree with @ericboehs , you can't use lambda with Marshal.dump like this:
```
[2] pry(main)> Marshal.dump(lambda { puts '123' })
TypeError: no _dump_data is defined for class Proc
```
In normal case you can use some hooks, which are valid:
```ruby
serialized_block = %{lambda { puts '123' }}
Marshal.dump(eval(serialized_block).call)
```
or you can use sourcify gem (use only latest 0.6.0.rc4 gem )
```ruby
require 'sourcify'
l = lambda { puts '123' }
serialized_block = l.to_source
Marshal.dump(eval(serialized_block).call)
```

But in this case, i don't find the solution, all attempts are failed because i get error, which can't debug
```
Exception: SystemStackError: stack level too deep
```
It looks, here
```ruby
 Response.new(self, last_response, lambda { parse_response(body) }, body: body)
```
parse_response have a complicated stack (Request.parse_response, Parser.call etc)
@jnunemaker , what you think about it ? Do you have any ideas how to solve this issue?
Also, may be helpful for somebody will be 'updated test' in first message for current master: 
```ruby
require File.expand_path(File.join(File.dirname(__FILE__), '..', 'spec_helper'))

RSpec.describe "HTTParty marshal" do  
  it "should be serailizable (for memcaching)" do 
[...]

PR #714, opened 2020-10-31
Fix #713

### LINKED ISSUE #713: Marshal.dump fails on response objects when request option `:logger` is set, or `:parser` is a proc
I discovered this issue while trying to cache responses from an API. The issue occurs when one calls `Marshal.dump` on the response generated by `HTTParty.get` when the `:logger` option is set, or the `:parser` option points to a `Proc`. To reproduce the logger issue:

```ruby
require "httparty"
require "logger"

logger = Logger.new($stdout)
opts = { logger: logger }
res = HTTParty.get("https://google.com", **opts)
Marshal.dump(res) # This raises a TypeError (can't dump IO)
```

A similar issue occurs when trying to `Marshal.dump` a response that was generated when the `:parser` option was set and points to a proc. To reproduce this issue:

```ruby


require "httparty"

proc_parser = proc { |body| body }
opts = { parser: proc_parser }
res = HTTParty.get("https://google.com", **opts)
Marshal.dump(res) # This raises a TypeError (no _dump_data is defined for class Proc)
```

I have a PR incoming to fix these issues, pending some discussion and your approval.

baberthal:
This PR fixes the issue(s) described in #713. To fix the logger option
issue, I simply delete `Request.options[:logger]` when dumping. To fix
the proc parser issue, I delete `Request.options[:parser]` if and only
if `Request.options[:logger]` is a proc. If `Request.options[:logger]`
is a regular class, `Marshal.dump` should proceed as normal. This should
not affect the `Marshal.dump` behavior described in issue #143 and fixed
by PR #618.

I have added a feature spec to make sure marshalling the request
works as intended, as well as a unit test to ensure
`Marshal.load(Marshal.dump(req))` works as it should.

I would love to hear your feedback on this. I'm not crazy about simply deleting the `:logger` and `:parser` options, but I'm not sure how else to handle it. Thanks in advance.

### DIFF
--- a/features/steps/httparty_steps.rb
+++ b/features/steps/httparty_steps.rb
@@ -15,6 +15,15 @@ When /^I set my HTTParty header '(.*)' to value '(.*)'$/ do |name, value|
   @request_options[:headers][name] = value
 end
 
+When /I set my HTTParty logger option/ do
+  # TODO: make the IO something portable
+  @request_options[:logger] = Logger.new("/dev/null")
+end
+
+When /I set my HTTParty parser option to a proc/ do
+  @request_options[:parser] = proc { |body| body }
+end
+
 When /I call HTTParty#get with '(.*)'$/ do |url|
   begin
     @response_from_httparty = HTTParty.get("http://#{@host_and_port}#{url}", @request_options)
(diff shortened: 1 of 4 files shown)


jnunemaker:
@baberthal thanks! This looks good to me. @TheSmartnik any thoughts since you worked on #618?

PR #716, opened 2020-12-28
Fix Amazon search example

xymbol:
It moves the auth param to an instance variable. Also, it fixes how Active
Support core extensions are loaded.

https://guides.rubyonrails.org/active_support_core_extensions.html#cherry-picking-a-definition


jnunemaker:
@xymbol thanks for caring!

PR #721, opened 2021-02-09
Fix method call in spec for Ruby 3.0.0

bfad:
Ruby 3.0.0 requires transforming a hash to named parameters for method calls. This fix gets all tests to pass for Ruby 3.0.0 as well as older versions of Ruby (I tested 2.6.3).


TheSmartnik:
Thank you!

PR #724, opened 2021-03-10
Add RubyDoc link to readme

timwis:
Hi 👋 I always struggle to remember things like "how do I check the response status again?" and have to poke through the examples directory in hopes there might be something relevant. I thought that linking to the RubyDoc might help.


jnunemaker:
😂

PR #729, opened 2021-05-25
Better handling of Accept-Encoding / Content-Encoding decompression (fixes #562)

### LINKED ISSUE #562: GZip content not automatically being deflated
When making a HTTParty GZipped request with `Accept-Encoding`  set to `gzip, deflate` I would expect HTTParty to automatically un-GZip the response. What I am seeing is content which is still GZipped.

## To reproduce

Create HTTParty get request with  `Accept-Encoding`  set to `gzip, deflate` as below 
```ruby
r = HTTParty.get("https://example.com", {headers: { "Accept-Encoding" => "gzip, deflate"}})
r.body => # Still GZipped content
```

## Expected result 
```ruby
r = HTTParty.get("https://example.com", {headers: { "Accept-Encoding" => "gzip, deflate"}})
r.body => # Plain text content
```

This is related to post I created on https://github.com/campaignmonitor/createsend-ruby/issues/58 where they are doing the above.

*Edit:* The response body has to be over 1KB for the compressions to kick in. Which is where the bug appears

johnnyshields:
### Summary

Better handling of `Accept-Encoding` / `Content-Encoding` decompression:
1. By default, let `Net::HTTP` always do decompression, if it can.
2. Add support to decompress `"br"` ([Brotli](https://en.wikipedia.org/wiki/Brotli)) and `"compress"` (LZW) if the required gems are present. These are not supported by `Net::HTTP`
3. Add `:skip_decompression` option to disable both Net::HTTP and HTTParty decompression (i.e. 1 and 2 above)
4. Add "HTTP Compression" section to docs/README.md

This fixes #562 and reverts some `Accept-Encoding` behavior in v0.18.x back to how it was in v0.17.3 and prior.

### More Background

PR #678, released in v0.18.0, introduced undesirable behavior which broke my app in production.

* In v0.17.3 and prior, you could set the `Accept-Encoding` header (e.g. `gzip`, etc.) in a request, and `Net::HTTP` would automatically handle decompression the response body in most cases.
* In v0.18.0+, as a result of PR #678, when setting `Accept-Encoding`, `Net::HTTP` no longer performs automatic decompression. This means your `response.body` will be a gzipped byte-string, and `JSON.parse(request.body)` will fail. In my case, it actually failed with a process segfault because I use a C-based native JSON parsing lib called [Oj](https://github.com/ohler55/oj).

### The Way Forward

The "happy path" for the 99.9% majority of users to [party on](https://www.youtube.com/watch?v=MA5rg_LIcWY) is to have Net::HTTP auto-decompress whenever it can, even if you set the `Accept-Encoding` header. The 0.1% of purist users out there can use the new `:skip_decompression` option which disables auto-decompression in all cases, meaning you'll always get the raw response and the `Content-Encoding` header so you can roll your own decompression.

This change is fully backward compatible with versions prior to v0.18.0 (i.e. it fixes the 0.17.3 --> 0.18.0 breakage), and mostly compatible with v0.18.0+, with the caveat that users using `Accept-Encoding` may get an uncompressed response when they previous got a compressed one. It should be released as a minor bump `0.19.0`

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -104,3 +104,68 @@ class Client
   end
 end
 ```
+
+### HTTP Compression
+
+The `Accept-Encoding` request header and `Content-Encoding` response header
+are used to control compression (gzip, etc.) over the wire. Refer to
+[RFC-2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html) for details.
+(For clarity: these headers are **not** used for character encoding i.e. `utf-8`
+which is specified in the `Accept` and `Content-Type` headers.)
+
+Unless you have specific requirements otherwise, we recommend to **not** set
+set the `Accept-Encoding` header on HTTParty requests. In this case, `Net::HTTP`
+will set a sensible default compression scheme and automatically decompress the response.
+
+If you explicitly set `Accept-Encoding`, there be dragons:
+
+* If the HTTP response `Content-Encoding` received on the wire is `gzip` or `deflate`,
+  `Net::HTTP` will automatically decompress it, and will omit `Content-Encoding`
+  from your `HTTParty::Response` headers.
+
+* For encodings `br` (Brotli) or `compress` (LZW), HTTParty will automatically
+  decompress if you include the `brotli` or `ruby-lzws` gems respectively into your project.
+  **Warning:** Support for these encodings is experimental and not fully battle-tested.
+  Similar to above, if decompression succeeds, `Content-Encoding` will be omitted
+  from your `HTTParty::Response` headers.
+
+* For other encodings, `HTTParty::Response#body` will return the raw uncompressed byte string,
+  and you'll need to inspect the `Content-Encoding` response header and decompress it yourself.
+  In this case, `HTTParty::Response#parsed_response` will be `nil`.
+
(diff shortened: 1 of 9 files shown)


johnnyshields:
@jnunemaker appreciate if you can review this.

jnunemaker:
I think what you are saying makes sense to me. I don't know my way around this area much. But I'm inclined to merge because you have been so thorough and obviously have more experience than me here. 

That said, any ideas why the ssl certificate spec is failing? It does not on master but does on this branch. I even tried locally and got the same result.

johnnyshields:
@jnunemaker which spec specifically is failing? Could it be related to the fixes in this PR: https://github.com/jnunemaker/httparty/pull/702

jnunemaker:
@johnnyshields hmm, I don't think so because master is passing on GitHub and my local machine and that PR isn't merged.

johnnyshields:
@jnunemaker I fixed the issue. It was an issue with RSpec mocks, not with the actual SSL functionality.

johnnyshields:
@jnunemaker any update on this?

jnunemaker:
@johnnyshields thanks again!

johnnyshields:
@jnunemaker thank you. Can you cut this as a new release 0.19.0?

PR #732, opened 2021-09-10
CI: use ruby/setup-ruby + bundler-cache

olleolleolle:
This PR leverages the bundler-cache feature of the Ruby-maintained Action setup-ruby.

[See ruby/setup-ruby](https://github.com/ruby/setup-ruby/blob/master/action.yml) for more information.


jnunemaker:
Thanks!

PR #737, opened 2021-09-29
CI: Add "3.0" to build matrix

olleolleolle:
This is a very small change, add Ruby 3.0 to build matrix.

...but which avoids a Float-to-String loss of characters.

An example (not from this project) of what it can look like before this change, in this picture:

bild

Read more details, if you like: https://github.com/actions/runner/issues/849


jnunemaker:
Thanks for caring!

PR #748, opened 2022-02-22
Add Thread Safety checks and patches using `rubocop` and `rubocop-thread_safety`

tcannonfodder:
1. Added https://github.com/covermymeds/rubocop-thread_safety to help identify thread safety issues
2. Updated `.rubocop*` to remove errors & warnings so that Rubocop would run
3. Enabled `ThreadSafety` checks for Rubocop, ignoring `spec/**/*`
4. Added a step in the Github action to run the Rubocop thread safety check
5. Updated `ConnectionAdapter.default_cert_store` & `Logger.formatters` to use thread variables

I tried refactoring `ModuleInheritableAttributes` to try not to use instance variables on the class, but couldn't figure it out. Any help would be super appreciated!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,3 +23,5 @@ jobs:
           bundler-cache: true # Run "bundle install", and cache the result automatically.
       - name: Run Rake
         run: bundle exec rake
+      - name: Run ThreadSafety Check
+        run: bundle exec rubocop --only ThreadSafety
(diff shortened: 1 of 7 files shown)


jnunemaker:
> I tried refactoring ModuleInheritableAttributes to try not to use instance variables on the class, but couldn't figure it out. 

We probably just need to see what rails/active support or similar do. 

Closing as stale due totally to my fault. But open to any future work on this.

PR #749, opened 2022-02-24
Support extra chain cert

wxwang33:
**Summary**

This change is to add extra chain cert support:

Currently, httparty does not support having multiple certs in the `pem` option. The only way to pass in multiple certs is to use a ca_file and this pull request will add support to passing multiple certs without the need of a file.


jnunemaker:
Forgive my ignorance, because I know nothing about this, but is it a common thing or specification to do this?

jnunemaker:
@wxwang33 the only thing we need on this now is specs. Can you get me an example file that has two certs like this? I can help wire up the specs for it.

jnunemaker:
Closing as stale but if someone wants to help get some specs for this I'll merge it.

PR #752, opened 2022-06-26
Support zstd decompression

johnnyshields:
Follows same implementation of how Brotli and LZW are done today.


jnunemaker:
@johnnyshields hey thanks! Appreciate the addition. Sorry I was out for an unusually long vacation but I'm back so I'm merging.

PR #760, opened 2022-08-18
Lazy load some gems

orgads:
Only load them when they are used.

### DIFF
--- a/features/steps/mongrel_helper.rb
+++ b/features/steps/mongrel_helper.rb
@@ -30,6 +30,7 @@ class DeflateHandler < BasicMongrelHandler
       reply_with(response, 406, 'No deflate accept encoding found in request')
     else
       response.start do |head, body|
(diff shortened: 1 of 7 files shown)


orgads:
Rebased.

jaredbeck:
Cross-posting this here, in case it helps someone:

> On a related note, removing the top-level requires in https://github.com/jnunemaker/httparty/pull/760 may cause issues for some users who fail to require their own dependencies (speaking for a friend 😄 ) For example, if my friend's codebase requires CSV but they fail to specify that dependency in any way, then they will see `NameError: uninitialized constant .. CSV` after updating httparty to 0.22. This is not a breaking change in httparty, but you might choose, in an abundance of compassion, to mention it in the release notes. 😆
> https://github.com/jnunemaker/httparty/issues/801#issuecomment-2150822317

PR #762, opened 2022-09-15
added response time logging

sreejithgp:
Thought it would be nice to show the response time for a request in the logs


jnunemaker:
Sorry for the late reply. I'd just make a customer formatter for this if you want it. Unless you have a link showing that the apache log format includes response time. If so, drop it here and I'll re-open this -- though I would have some minor changes.

Issue #743, opened 2021-12-03
Please specify in documentation to convert body in json


stefanosalvucci:
I've spent one hour figuring out how to proper send a POST request, ended up on https://stackoverflow.com/questions/7455744/post-json-to-api-using-rails-and-httparty which helped me a lot in adding **.to_json** to the body hash

@result = HTTParty.post(@urlstring_to_post.to_str, 
    :body => { :subject => 'This is the screen name', 
               :issue_type => 'Application Problem', 
               :status => 'Open', 
               :priority => 'Normal', 
               :description => 'This is the description for the problem'
             }.**to_json**,
    :headers => { 'Content-Type' => 'application/json' } )

JonMidhir:
This is probably common enough that it could be detailed in the docs. @jnunemaker could you review https://github.com/jnunemaker/httparty/pull/765 please?

jnunemaker:
Merged the documentation update for this.

PR #765, opened 2022-10-07
Add docs on JSON post, put and patch requests

JonMidhir:
Addresses https://github.com/jnunemaker/httparty/issues/743

This change adds docs regarding how to send data in JSON format. Users are often<sup>[1][2]</sup> on how to format JSON data to use in POST, PUT and PATCH requests. This adds simple examples on how to do so.

1: https://stackoverflow.com/questions/7455744/post-json-to-api-using-rails-and-httparty
2: https://github.com/jnunemaker/httparty/issues/743

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -14,6 +14,20 @@ response = HTTParty.get('http://example.com', format: :plain)
 JSON.parse response, symbolize_names: true
 ```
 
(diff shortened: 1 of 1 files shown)


jnunemaker:
@JonMidhir thanks!

PR #769, opened 2022-11-10
Suggestion: Replace `mime-types` with `mini_mime` for content type lookup

carlosantoniodasilva:
`mini_mime` is a minimal mime type library that's more performant and less memory hungry.
https://github.com/discourse/mini_mime

It has replaced `mime-types` in the `mail` gem: (which is a dependency of `actionmailer`, and by extension, `rails`)
https://github.com/mikel/mail/pull/1059

As well as `capybara`:
https://github.com/teamcapybara/capybara/pull/1884

Which also means using it as a dependency of `httparty` would be able to reuse the same dependency that should be already available in most Rails apps, instead of pulling in an extra `mime-types` dependency.

The change in code is pretty straightforward, the same one made by the capybara PR linked above.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 ## unreleased
 
 * Fix request marshaling (https://github.com/jnunemaker/httparty/pull/767)
+* [Replace `mime-types` with `mini_mime`](https://github.com/jnunemaker/httparty/pull/769)
 
 ## 0.20.0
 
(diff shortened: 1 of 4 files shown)


jnunemaker:
This is great! I'm sorry I didn't see it until now. Not sure what the deal is but I swear I'm not getting some notifications.

PR #776, opened 2023-01-11
Avoid anonymous eval

casperisfine:
This makes it hard to track down code when profiling etc.


jnunemaker:
Sorry I missed this! Feels like I'm not getting some notifications...

PR #783, opened 2023-05-02
Clear body when redirecting to a GET 

rhett-inbox:
Problem: Ran into a web application firewall that was blocking requests because the originating POST body was retained and transmitted upon redirection. 

Solution: When redirecting, clear the body if the subsequent request is a GET

Note: Also extracted a method to keep the handle_response method more concise and cohesive IMHO

### DIFF
--- a/lib/httparty/request.rb
+++ b/lib/httparty/request.rb
@@ -295,24 +295,7 @@ module HTTParty
 
     def handle_response(raw_body, &block)
       if response_redirects?
-        options[:limit] -= 1
-        if options[:logger]
-          logger = HTTParty::Logger.build(options[:logger], options[:log_level], options[:log_format])
-          logger.format(self, last_response)
-        end
-        self.path = last_response['location']
-        self.redirect = true
-        if last_response.class == Net::HTTPSeeOther
-          unless options[:maintain_method_across_redirects] && options[:resend_on_redirect]
-            self.http_method = Net::HTTP::Get
-          end
-        elsif last_response.code != '307' && last_response.code != '308'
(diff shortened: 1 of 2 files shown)


jnunemaker:
Hi! Sorry I missed this. Is this what curl and other things do? Or is it a standard somewhere? I never know what to do in situations like this since I'm not sure if that is what everyone would want it to do or if it should be an option. Any insight would help.

rhett-inbox:
> Hi! Sorry I missed this. Is this what curl and other things do? Or is it a standard somewhere? I never know what to do in situations like this since I'm not sure if that is what everyone would want it to do or if it should be an option. Any insight would help.

hey @jnunemaker , no prob. Yes I believe it is a de facto standard.  Here's one [page ](https://peterdaugaardrasmussen.com/2020/05/09/how-to-redirect-http-put-or-post-requests/)that describes the behavior which is much easier to digest. I think the fact that the AWS infra was blocking the requests is additional evidence. And then when I was troubleshooting I wasn't able to repro in Postman because it was dropping the body too. 

I hope that mix of empirical and "canonical" info helps. I sympathize with the quandry, but I do think that this is the right choice for gem. 

Happy to explore the convo further.

shashankjo (approved):

chiraggshah:
@jnunemaker : cURL and Postman does it too.

captn3m0:
Facing this at https://github.com/captn3m0/outliner/pull/9 (CloudFront WAF by default blocks GET requests with a body). +1 to merging this, as current behaviour goes against both existing-usage and spec.

PR #800, opened 2024-05-03
Add csv to Gemfile

ltickett:
It will no longer be available as a system package in Ruby 3.4.0


ltickett:
Looks like this was already fixed in 0.22.0 🚀

PR #802, opened 2024-07-16
Add support for including extra_chain_cert with p12

nparker-tc:
## Summary of feature
We ran into an integration where the partner required the full certificate chain included with SSL requests.  Adding this support in HTTParty allows the option to include the full certificate chain utilizing the existing data from the p12 certificate.  This feature has been included in `Net::HTTP` since Ruby **3.0**. I have added a safety check to prevent compatibility issues to ensure it is only accessible when the `extra_chain_cert=` method is available on the Net::HTTP instance.

## Techinical Details
- Added option to determine if the full certificate chain should be included.  Defaults to false
- Add support in connection_adapter to include the full certificate chain if the option is enabled, p12 is being used, and Net::HTTP supports including the certificate chain
- Added specs for both new methods

### DIFF
--- a/lib/httparty.rb
+++ b/lib/httparty.rb
@@ -353,6 +353,11 @@ module HTTParty
       default_options[:p12_password] = password
     end
 
+    # Allow using a full certificate chain (http.extra_chain_cert)
+    def extra_chain_cert(value = false)
+      default_options[:extra_chain_cert] = value
+    end
+
     # Override the way query strings are normalized.
     # Helpful for overriding the default rails normalization of Array queries.
     #
--- a/lib/httparty/connection_adapter.rb
+++ b/lib/httparty/connection_adapter.rb
@@ -233,7 +233,8 @@ module HTTParty
         end
 
         # Include full certificate chain
-        if options[:extra_chain_cert] && options[:p12]
+        # Only Ruby 3.0+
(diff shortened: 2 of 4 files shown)


jnunemaker:
Looks good. I've not used this and know nothing about it but I'll merge after CI passes.

nparker-tc:
@jnunemaker Wow thanks for the quick response!  I will re-try locally the 2.7 build and get it fixed. Appreciate it!

nparker-tc:
@jnunemaker 

The previous issue was with the way the specs were written. It would run the expectation regardless of whether it was supported on that ruby version or not.  In the new specs, both scenarios are checked for the supported and unsupported ruby versions and validate the behavior for both.

### Ruby 3.0 build:
image

### Ruby 2.7.8 build:
image

jnunemaker:
@nparker-tc 2.7 is failing. Any ideas why? I didn't look closely.

nparker-tc:
> @nparker-tc 2.7 is failing. Any ideas why? I didn't look closely.

Is it still?  I think it is currently awaiting on a workflow approval for a new build with the updated specs.  It failed previously based on the way the specs were written. More details are in this comment: https://github.com/jnunemaker/httparty/pull/802#issuecomment-2231530771

nparker-tc:
@jnunemaker Please let me know if you have any questions on this. It should be good to go

PR #803, opened 2024-07-21
docs: replace master branch to main for better view

bestony:


olleolleolle (commented):
This is a good change, and with it, the user navigating to those links won't see a blue bar at the top of the page saying "Branch master has changed name to main."

PR #813, opened 2025-04-10
Add support for net/http continue_timeout and ssl_timeout

lukes:
This change adds support for net/http's `continue_timeout` and `ssl_timeout` values.

If this direction is what we want, I'll add some tests.


jnunemaker:
Works for me! Happy to merge with some tests. Just search for the other timeout tests and duplicate likely.

PR #825, opened 2025-11-06
feat: set default content type when processing body

zygzagZ:
httparty does properly set multipart content type when dealing with files or multipart url bodies. There was no default for urlencoded.

net-http 0.7.0 stops defaulting content type to x-www-form-urlencoded ([net-http PR #207](https://github.com/ruby/net-http/pull/207), issue [#205](https://github.com/ruby/net-http/issues/205)). HTTParty encodes body with urlencode by default, but does not set content-type. HTTParty's default behavior of urlencoding needs to set content type. 

changes in this pr:
- default content type is set when implicitly encoding body as urlencode, as before with multipart
- content type default is not used when set or when using custom query_string_normalizer.

If you use custom query_string_normalizer, return content_type as a second argument or pass a `Content-Type` header

### DIFF
--- a/lib/httparty/request/body.rb
+++ b/lib/httparty/request/body.rb
@@ -16,7 +16,12 @@ module HTTParty
 
       def call
         if params.respond_to?(:to_hash)
-          multipart? ? generate_multipart : normalize_query(params)
+          if multipart?
+            @content_type = "multipart/form-data; boundary=#{boundary}"
+            generate_multipart
+          else
+            normalize_query_and_set_content_type(params)
+          end
         else
           params
         end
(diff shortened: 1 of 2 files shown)


lib/httparty/request/body.rb
```diff
@@ -30,6 +35,13 @@ def multipart?
         params.respond_to?(:to_hash) && (force_multipart || has_file?(params))
       end
 
+      def content_type
```
zygzagZ: it acts as a getter when set, and tries to calc when body is not yet evaluated.

lib/httparty/request/body.rb
```diff
@@ -16,7 +16,12 @@ def initialize(params, query_string_normalizer: nil, force_multipart: false)
 
       def call
         if params.respond_to?(:to_hash)
-          multipart? ? generate_multipart : normalize_query(params)
+          if multipart?
+            @content_type = "multipart/form-data; boundary=#{boundary}"
+            generate_multipart
+          else
+            normalize_query_and_set_content_type(params)
+          end
```
zygzagZ: useless change

zygzagZ: cus generate_multipiart sets `@content_type`, too

marelons1337:
Can we get this merged? As it does actually seem like a better idea than #828

Issue #832, opened 2025-12-29
Disable stream_body by defaut


igorkasyanchuk:
Hello, 

In our app, having streaming enabled by default caused an issue with 3rd party service.

We started to receive a 400 error.

Once I enabled stream_body: false, everything started to work as before.

Probably this should be enabled when needed.

jnunemaker:
Oh weird. I didn’t mean to change that default. Must have let it slip. I’ll
check and tweak.

PR #833, opened 2025-12-31
Add trailing clrf when building header for non-file parts.

serixscorpio:
When building a `StreamingMultipartBody`, the streaming body should produce same content as non-streamining.  The following example shows the header for the `first_name` part (a non-file part) is missing a clrf before the part's content starts.  I think this explains the 400 error described in #832 .

==== Before: no extra blank line between `"first_name"` and `John`  ====
```
--test-boundary-123
Content-Disposition: form-data; name="avatar"; filename="tiny.gif" Content-Type: image/gif

GIF89a,;
--test-boundary-123
Content-Disposition: form-data; name="first_name"
John
--test-boundary-123
```

The fix adds a trailing clrf to a non-file part's header:

==== After: notice the extra blank line between `"first_name"` and `John` ====
```
--test-boundary-123
Content-Disposition: form-data; name="avatar"; filename="tiny.gif" Content-Type: image/gif

GIF89a,;
--test-boundary-123
Content-Disposition: form-data; name="first_name"

John
--test-boundary-123
```

The Before/After outputs are generated by adding `puts streaming_content` to the updated `streaming body produces same content as non-streaming` test.

### DIFF
--- a/lib/httparty/request/streaming_multipart_body.rb
+++ b/lib/httparty/request/streaming_multipart_body.rb
@@ -135,6 +135,8 @@ module HTTParty
           header << %(; filename="#{file_name(value).gsub(/["\r\n]/, replacement_table)}").b
           header << NEWLINE.b
           header << "Content-Type: #{content_type(value)}#{NEWLINE}".b
+        else
+          header << NEWLINE.b
         end
         header << NEWLINE.b
         header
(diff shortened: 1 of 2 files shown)


serixscorpio:
The CI build for 3.4 failed due to (I think) bad timing.  The test diff shows the timestamp to be off by 1 second.

image

Maybe rerun the CI to see if we get a green?

jtgrenz:
related https://github.com/jnunemaker/httparty/issues/834

jnunemaker:
It's out.

PR #838, opened 2026-06-12
Use MultiXML constant to fix deprecation warning from multi_xml 0.9+

robzolkos:
multi_xml 0.9.0 renamed the top-level constant from \`MultiXml\` to \`MultiXML\` and made the old name a deprecated shim that emits a warning on every method call:

\`\`\`
The MultiXml constant is deprecated and will be removed in v1.0. Use MultiXML instead.
\`\`\`

This uses \`defined?(MultiXML)\` to prefer the new constant when available (multi_xml >= 0.9), falling back to \`MultiXml\` on older versions. No minimum version bump required — the fix is fully backward compatible.

### DIFF
--- a/lib/httparty/parser.rb
+++ b/lib/httparty/parser.rb
@@ -119,7 +119,7 @@ module HTTParty
 
     def xml
       require 'multi_xml'
-      MultiXml.parse(body)
+      (defined?(MultiXML) ? MultiXML : MultiXml).parse(body)
     end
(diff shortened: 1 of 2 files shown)


pioz:
Yeah, please merge and push a new minor release! 🙏🏻

jnunemaker:
Sorry I missed this. Getting it working now.

PR #844, opened 2026-09-08
Remove JSON `:quirks_mode` method from `:json` method

rargumedo-mccorvey:
As of JSON 3.0.0, the `quirks_mode` method has been removed and no longer supported. The original behavior has the following code that was hardcoded:

```ruby
JSON.parse(body, :quirks_mode => true, :allow_nan => true)
```

so on json >= 3.0 every JSON response HTTParty tries to parse raises:

```
ArgumentError: unknown keyword: quirks_mode
```

This breaks any app that upgrades to json 3.x while still using HTTParty for JSON APIs.

### DIFF
--- a/lib/httparty/parser.rb
+++ b/lib/httparty/parser.rb
@@ -126,7 +126,7 @@ module HTTParty
 
     def json
       require 'json'
-      JSON.parse(body, :quirks_mode => true, :allow_nan => true)
+      JSON.parse(body, :allow_nan => true)
     end
 
     def csv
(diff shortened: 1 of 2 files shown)


rargumedo-mccorvey:
## Related

This was found while investigating a `json` 3.0 upgrade for another Rails app — HTTParty was one of two blockers found (the other is inside `ActiveSupport::JSON.decode` itself, already fixed on Rails `main` in [rails/rails@cc07aa3](https://github.com/rails/rails/commit/cc07aa3153e3bd47151b6bbaa0f39347b0feb4bc), not yet in a released gem).

